PR #3, opened 2026-05-28
Various madmatrix improvements and bugfixes

theoheimel:
GPU:
- reduce number of `cudaMallocAsync` calls in UMAMI interface
- allow for fully asynchronous calls of `sigmaKin` without device or stream synchronization
- allow for parallelization over helicities in a single kernel launch instead of using CUDA streams

SIMD:
- use one flavor index per SIMD vector instead of per batch
- reorder events in UMAMI such that the flavor index is the same for each vector


madgraph/iolibs/template_files/madmatrix/umami.cc
```diff
@@ -336,22 +330,38 @@ extern "C"
...
+        {reinterpret_cast<void**>(&numerators), rounded_count * CPPProcess::ndiagrams * CPPProcess::ncomb * sizeof( fptype )},
+        {reinterpret_cast<void**>(&denominators), rounded_count * CPPProcess::ncomb * sizeof( fptype )},
+        {reinterpret_cast<void**>(&helicity_index), rounded_count * sizeof( int )},
+        {reinterpret_cast<void**>(&color_index), rounded_count * sizeof( int )},
+        {reinterpret_cast<void**>(&ghel_matrix_elements), rounded_count * CPPProcess::ncomb * sizeof( fptype )},
+        {reinterpret_cast<void**>(&ghel_jamps), rounded_count * CPPProcess::ncomb * CPPProcess::ncolor * mgOnGpu::nx2 * sizeof( fptype )},
+    }};
+    std::size_t total_size = 0;
+    for (auto [ptr, size] : ptrs_and_sizes) {
+        std::size_t aligned_size = (size + 7) / 8 * 8;
```
Qubitol: Maybe naive question: this is used to round to the closest multiple of 8 because 8 is `sizeof(double)`? If that's the case, I suggest 2 things: to use `fptype` and not `double`, and I propose to make the numbers explicit by doing like `const std::size_t ROUND = sizeof(fptype); std::size_t aligned_size = (size + ROUND - 1) / ROUND * ROUND;`, which helps a bit for readibility while constants are optimised out.

madgraph/iolibs/template_files/madmatrix/umami.cc
```diff
@@ -503,35 +541,71 @@ extern "C"
...
-      if( m2_out != nullptr )
-      {
-        m2_out[i_event + offset] = matrix_elements[i_event];
-      }
-      if( amp2_out != nullptr )
+      for( std::size_t i_event = 0; i_event < count; ++i_event )
       {
-        for( std::size_t i_diag = 0; i_diag < CPPProcess::ndiagrams; ++i_diag )
+        std::size_t i_sorted = permutation[i_event];
+        std::size_t page_size = MemoryAccessMomentaBase::neppM;
```
Qubitol, line 564: Maybe let's call it `simd_vector_size` instead of `page_size` to avoid confusion, since we have already _pages_ to handle the memory. I guess it can be made constant as well.

madgraph/iolibs/template_files/madmatrix/umami.cc
```diff
@@ -503,35 +541,71 @@ extern "C"
...
+        std::size_t i_page = i_sorted / page_size;
+        std::size_t i_vector = i_sorted % page_size;
```
Qubitol:
```suggestion
        std::size_t i_vector = i_sorted % page_size;  // vector lane
```

Qubitol (changes requested):
I just added few comments on nomenclature, just to try to improve readibility from a not-expert of the code. But the logic looks good to me!

PR #5, opened 2026-06-11
Introduction of massive RAMBO in standalone_mg7.

stloufra:
For now Host only, if MGONGPUCPP_GPUIMPL momenta buffer copy from host to device
Uses its own internal RNG - to be decided if to be kept
New flag in check_sa perf -r [c|ml] (classic or massless) 
Default for both modes is massive (classic) 
New get_meta option in umami to retrieve masses from CPPProcess


madgraph/iolibs/template_files/madmatrix/check_sa.cc
```diff
@@ -76,19 +76,22 @@ namespace
...
 
+  enum RamboType { RAMBO_CLASSIC, RAMBO_MASSLESS };
+
   int usage( const char* argv0, int ret = 1 )
   {
     std::cout
       << "Usage:\n"
       << "  " << argv0 << " [matrix] [-v|--verbose]\n"
-      << "  " << argv0 << " perf [-v|--verbose] [-f|--flavor <int>]"
+      << "  " << argv0 << " perf [-v|--verbose] [-f|--flavor <int>] [-r|--rambo [c]|ml]"
```
Qubitol: Given the default is *classic* and it makes total sense, I would just add a flag `--rambo-massless` which works as a switch.

madgraph/iolibs/template_files/madmatrix/RamboSamplingKernels.cc
```diff
@@ -11,6 +11,7 @@
 #include "MemoryAccessRandomNumbers.h"
 #include "MemoryAccessWeights.h"
 #include "MemoryBuffers.h"
+#include "classic_rambo.h" // inline classic (massive) RAMBO, ported from standalone_cpp
```
Qubitol: Maybe we can call it just `rambo.h` given it is the sane default, while the other becomes `massless_rambo.h`?

Qubitol (changes requested):
Just some general comments on namings. It looks good for the rest. Given now we have a proper rambo implementation, I would make it default all the way through (including the name) and treat the massless rambo as a special case.

Qubitol (approved):

PR #7, opened 2026-06-17
Merging with aloha_obj_merged + standalone_gpu in check language + masking for standalone_mg7

oliviermattelaer:


oliviermattelaer:
@Qubitol @theoheimel Would you aggree to merge this one?
I'm pushing because this will simplify our life with the handling of cache between branch.
If you prefer to wait, I can create a simpler PR, where I only update the cache handling such that this can be merge into main quickly (which might be cleaner)

Qubitol:
Yes, we can merge, I checked few decay chains and matrix element is ok for both fortran and MadMatrix.

theoheimel:
Yes, let's merge it. I don't see any problems with that on the mg7/madspace side of things

oliviermattelaer:
Thanks a lot !!!

PR #8, opened 2026-06-23
merge with fix for goodhel at the fortran level, cleaner fortran implementation for the standalone part

oliviermattelaer:
Crucially, this spot an issue with the helicity filtering within standalone_mg7 and added a CI to check it


Qubitol (approved):
No problem for me here. The test on the helicity filter looks good. The rest is mostly on the Fortran side, and I trust you on that.
Good to merge if feature-complete.

PR #10, opened 2026-06-23
Feat madspace ramon

ramonpeter:
Features added:
- Cuts are directly piped to the 2->2 and 2->3 mappings to improve cutting efficiency.
- ColorOrderedMapping is now embedded in the top-level PhaseSpaceMapping class and is being selected as t-channel strategy, taking specific color-order to build its mapping
- Splitting of discrete and continuous random numbers. Relevant for the ColorOrderedMapping which takes a discrete index for each 2->3 mapping


madspace/src/phasespace/phasespace.cpp
```diff
@@ -178,26 +229,86 @@ PhaseSpaceMapping::PhaseSpaceMapping(
...
+            // children carry no pairwise cut.
+            const auto& out_idx = topology.outgoing_indices();
+            std::vector<std::size_t> child_to_out(
+                topology.decays().at(0).child_indices.size(),
+                std::numeric_limits<std::size_t>::max()
+            );
+            for (std::size_t a = 0;
+                 std::size_t cidx : topology.decays().at(0).child_indices) {
+                for (std::size_t p = 0; p < out_idx.size(); ++p) {
+                    if (out_idx[p] == cidx) {
```
theoheimel: Don't use unchecked array access `arr[i]` since this can lead to hard-to-find memory bugs. If possible, avoid indexing altogether using range-based for loops, potentially `zip`. Otherwise use checked access with `arr.at(i)`. Applies to a few different places in this pull request.

madspace/src/phasespace/two_particle.cpp
```diff
@@ -78,7 +78,9 @@ TwoToTwoParticleScattering::TwoToTwoParticleScattering(
          {"mass1", batch_float},
          {"mass2", batch_float}},
         {{"momentum1", batch_four_vec}, {"momentum2", batch_four_vec}},
-        {{"momentum_in1", batch_four_vec}, {"momentum_in2", batch_four_vec}}
+        {{"momentum_in1", batch_four_vec},
+         {"momentum_in2", batch_four_vec},
+         {"t_min_cut", batch_float}}
```
theoheimel: Not sure about this change to the interface. It makes the building block more complicated to use as you always need to pass the cut values as an extra input array. The input to `t_min_cut` appears to be constant in all places where it is used (I might be wrong here), so maybe a better solution is to pass it as an optional argument to the constructor of the building block and then choose between `t_inv_value_and_min_max` and `t_inv_value_and_min_max_cut` depending on whether it is zero or not.

madspace/src/phasespace/cuts.cpp
```diff
@@ -91,3 +91,56 @@ std::vector<double> Cuts::pt_min() const {
...
+        j -= 2;
+        if (i < n && j < n && item.min > m.at(i).at(j)) {
+            m.at(i).at(j) = item.min;
+            m.at(j).at(i) = item.min;
+        }
+    }
+    return m;
+}
+
+std::vector<std::vector<double>> Cuts::dr_min() const {
```
theoheimel: Factor out duplicated code into a function

theoheimel (changes requested):
See suggested changes

madspace/src/phasespace/phasespace.cpp
ramonpeter: Addressed in new commit

madspace/src/phasespace/two_particle.cpp
ramonpeter: Addressed in new commit

madspace/src/phasespace/cuts.cpp
ramonpeter: Addressed in new commit

ramonpeter:
test_pdf does not work anymore! probably linked to some changes I merged from main!
@theoheimel thats linked to some of your changes

PR #17, opened 2026-06-25
Fix stale flavortag on shared wavefunctions in HelasDiagram.check_flavor

oliviermattelaer:
## Problem

`HelasDiagram.check_flavor` returned an **order-dependent** answer for some diagrams. The reset loop only deleted the `flavortag` attribute from the wavefunctions *introduced* by the diagram (`self['wavefunctions']`). In a HELAS-optimised matrix element a diagram also **reuses** wavefunctions (and their whole sub-trees) introduced by earlier diagrams — those appear here only as mothers. Those shared mothers kept a stale `flavortag` computed for a *different* flavor, and `propagate_flavor_tag` silently reused it instead of recomputing.

Concrete symptom (reported by a colleague): for `q q > q q g` / `q q~ > q q~ g`, the t-channel-gluon diagrams where the external gluon is radiated off a final quark have their internal lines (t-channel gluon + internal quark) entirely reused from earlier diagrams. When checked *after* another flavor had been checked on the same matrix element, these diagrams were wrongly rejected for a valid **same-flavor** assignment (e.g. `d d > g d d` returned `False`).

## Fix

Clear `flavortag` over the **whole ancestor closure** of every wavefunction and amplitude (with a visited-set to stay O(n)). External leaves are re-tagged as before; cleared internal shared mothers are recomputed by `propagate_flavor_tag`.

This is the same bug class previously patched at the `check_flavor_for_all_diagrams` level; the fix here makes the lower-level `HelasDiagram.check_flavor` correct on its own.

## Tests

Added three regression test classes in `tests/unit_tests/core/test_helas_objects.py`:

| Class | Process |
|---|---|
| `TestDiagramFlavorCheck` | d d~ > d d~ |
| `TestDiagramFlavorCheckQQG` | q q > q q g |
| `TestDiagramFlavorCheckQQbarG` | q q~ > q q~ g |

Each pins that the t-channel gluon diagrams are valid for same-flavor assignments, that `check_flavor` genuinely discriminates mixed flavors, and that the matrix-element-level check conserves flavor.

## Verification

- Full `test_helas_objects.py`: **56/56 pass**
- With the fix reverted, `TestDiagramFlavorCheckQQG` fails 2 and `TestDiagramFlavorCheckQQbarG` fails 1 — confirming the tests genuinely guard the regression; all pass with the fix
- `TestFlavorCheck` (flavor grouping, process_checks): 4/4 pass
- `u q > Z u q` diagram trimming verified order-independent (`check_flavor_for_all_diagrams` unaffected)


oliviermattelaer:
@theoheimel 
This should fix your issue. (let's see the status of the CI before merging into main)

oliviermattelaer:
ok this is based on the wrong branch. let's cancel this PR and cherry-pick it on top of  main (and then redo a PR)

PR #18, opened 2026-06-25
Fix stale flavortag on shared wavefunctions in HelasDiagram.check_flavor

oliviermattelaer:
HelasDiagram.check_flavor only cleared the `flavortag` attribute on the wavefunctions *introduced* by the diagram (self['wavefunctions']). In a HELAS-optimised matrix element a diagram also reuses wavefunctions (and their sub-trees) introduced by earlier diagrams, which appear here only as mothers. Those shared mothers kept a stale flavortag computed for a different flavor, and propagate_flavor_tag silently reused it instead of recomputing it. This made check_flavor order-dependent: a diagram whose internal lines are all reused (e.g. the t-channel gluon diagrams of q q > q q g and q q~ > q q~ g where the external gluon is radiated off a final quark) was wrongly rejected for a valid same-flavor assignment when checked after another flavor.

Clear the flavortag over the whole ancestor closure of every wavefunction and amplitude instead. External leaves are re-tagged as before; cleared internal shared mothers are recomputed by propagate_flavor_tag. This is the same bug class previously patched at the check_flavor_for_all_diagrams level; the fix here makes the lower-level HelasDiagram.check_flavor correct on its own.

Add regression tests (TestDiagramFlavorCheck, TestDiagramFlavorCheckQQG, TestDiagramFlavorCheckQQbarG) covering d d~ > d d~, q q > q q g and q q~ > q q~ g; the latter two fail without this fix.

### DIFF
--- a/madgraph/core/helas_objects.py
+++ b/madgraph/core/helas_objects.py
@@ -3699,12 +3699,30 @@ class HelasDiagram(base_objects.PhysicsObject):
 
         #flavor_status = {}
         #pdg_for_number = {}
-        # remove the information from previous check
-        for wfct in self['wavefunctions'] + self['amplitudes']:
+        # remove the information from previous check.
+        # NB: self['wavefunctions'] only lists the wavefunctions *introduced*
+        # by this diagram.  In a HELAS-optimised matrix element a diagram also
+        # reuses wavefunctions (and their sub-trees) introduced by earlier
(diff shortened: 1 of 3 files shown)


oliviermattelaer:
@theoheimel do you want to do some test on this or we can merge it?

theoheimel (approved):
Looks good to me. I also tested a few processes and didn't see any with an empty list of active flavors.

PR #19, opened 2026-06-25
ci(mg7): build madspace once per workflow and cache the wheel

oliviermattelaer:
Add a single build_madspace job to acceptancetest_mg7.yml that compiles the madspace wheel and caches it under a content-addressed key (hashFiles('madspace/**')), so it is only rebuilt when the madspace sources change. All mg7 test jobs now depend on this job (needs: build_madspace) and install the cached wheel via the new install_madspace composite action (restore-only + pip install) instead of recompiling the C++/CUDA extension themselves. This guarantees exactly one madspace build per workflow run.

checkout_mg5 still installs madspace for now; removing that is a separate follow-up commit.


oliviermattelaer:
this still need to remove the old way to install madspace... 
This is postponed on purpose to check CI first, then remove it and resubmit and the failing jobs will need to be migrated between files (or guarded correctly to use the new install)

.github/actions/install_madspace/action.yml
```diff
@@ -0,0 +1,26 @@
...
+
+    - name: Restore madspace wheel
+      uses: actions/cache/restore@v5
+      with:
+        path: ~/.cache/madspace-wheel
+        key: madspace-${{ runner.os }}-py${{ env.MS_PY }}-${{ hashFiles('madspace/**') }}
+        fail-on-cache-miss: true
+
+    - name: Install madspace from wheel
+      run: python3 -m pip install ~/.cache/madspace-wheel/madspace-*.whl
```
theoheimel: generate_events will look for a madspace installation under `madspace/install` and otherwise start the installer. So this should be something like
```suggestion
      run: python3 -m pip install ~/.cache/madspace-wheel/madspace-*.whl --target=./madspace/install
```

.github/workflows/acceptancetest_mg7.yml
```diff
@@ -30,61 +30,107 @@ concurrency:
...
+
+      # Built from the in-tree source (scikit-build-core / cmake). Upgrade pip +
+      # packaging first: the runner ships an ancient pip whose build isolation
+      # does not provide packaging>=22, which scikit-build-core's vendored
+      # pyproject_metadata requires.
+      - name: Build madspace wheel (only on cache miss)
+        if: steps.cache-madspace.outputs.cache-hit != 'true'
+        run: |
+          cd $GITHUB_WORKSPACE
+          sudo apt-get install -y libopenblas-dev
```
theoheimel: openblas might lead to problems here since it is only installed when no cached madspace is found. We can either always install it or use the `-Ccmake.define.ENABLE_OPENBLAS=ON` of the madspace build to build it from source and link it statically.

theoheimel (changes requested):
See suggested changes

.github/actions/install_madspace/action.yml
oliviermattelaer: Then I have cache directly that directory which should be even more efficient

.github/workflows/acceptancetest_mg7.yml
oliviermattelaer: Done. Good suggestion

PR #22, opened 2026-06-30
Addition of reading of LHE file as source of phase space point and right amount of s.d. in LHE output

stloufra:
As asked on one of the previous developers meeting introduction of possibility of reading a LHE file in the standalone mode in the check_sa (only perf mode, automatic calc of iter from the grid * block size.

Also before, we outputted only 10-11 s.d. in the LHE file so corrected to 17.


Qubitol (approved):
Good for me. Thanks.

Issue #13, opened 2026-06-24
MG7 Hall of Shame


theoheimel:
Collect processes here that are currently not working, or give the wrong results.

**Careful**: the default scale choice in MG5 is not yet implemented in MG7, so if you want to compare cross section results you have to change the scaled to fixed (current default in MG7) or change it in both to, for instance, half transverse mass.

theoheimel:
# ✅ FIXED

In some cases the `export_mg7.py` script (https://github.com/MadGraphTeam/MadGraph7/blob/main/madgraph/iolibs/export_mg7.py#L89) yields an empty list for the available flavors of some channels. This leads to wrong integration results, especially when using MadNIS with trained channel weights. Found for example for `_quark _anti_quark > d d~` (second channel), `_quark _quark > t t~ _quark _quark` (5th and 6th channel), `_quark _quark > _quark _quark g` (4th and 5th channel).

-> partially fixed by https://github.com/MadGraphTeam/MadGraph7/pull/18 , however still wrong results for some diagrams. For `_quark _quark > g _quark _quark`, diagrams `1, 2, 9, 10`, flavor configurations of the type `xy>gxy` are wrongly ruled out.

theoheimel:
# ✅ FIXED 
Process: `p p > Z, Z > e+ e-` at fixed scale
MG5: `1422 +- 0.8357 pb`
MG7: `630.87(77) pb`

EDIT: this is related to cuts. In the decay syntax MG5 sets no cuts on leptons. After removing the cuts, MG7 produces the same result.

theoheimel:
# ✅ FIXED

Process `_quark _quark > _quark _quark g` yields `5.95(88)e+05 pb` in MG7, whereas `_quark _quark > g _quark _quark` yields `4.6054(66)e+05 pb`.

Potentially fixed by #7 , check again once merged

EDIT: #7 indeed fixed the discrepancy between the two orders with `6.4623(94)e+05` and `6.4592(96)e+05`, respectively. However it disagrees with the MG5 results `1.037e+06 +- 2719`

EDIT2: the remaining difference is from two bugs: the active_flavors bug (see above), and madspace not respecting that some flavor configs need a mirror process and others don't. Working on the latter in feat-flavorsym-fix.

oliviermattelaer:
# ✅ FIXED
g g > h QCD=0 [QCD]; output madevent; launch
[...]

PR #23, opened 2026-07-01
mg7 run_card.toml: editable RunCard representation + scriptable launch/install

oliviermattelaer:
## Summary

Give the default (mg7/madnis) generation mode a proper, editable `run_card.toml`, matching how the legacy `run_card.dat` works, and make the surrounding launch/install flow script-friendly.

Organised as four commits:

1. **RunCard-style representation of `run_card.toml`** — new `banner.RunCardMG7` (typed params, allowed values, comments, `user_set`, nested `card["section"]["key"]` access; free-form `[multiparticles]`/`[cuts]`/`[histograms]`). The card is generated from a placeholder template with process-dependent defaults (`create_default_for_process` + `input/.default_run_card_mg7.toml`) instead of copied verbatim, and the runtime consumes the class.
2. **Interactive card editing via `set`** — wire the TOML card into the "edit a card?" question so `set` edits it like `run_card.dat`, plus conveniences: `set <section>.<key>`, cut editing, energy units (→ GeV), arithmetic (`+ - * / **`), param_card mass refs (`mz`, `mh`, …), and shortcuts (`no_parton_cut`, `lhc`/`lep`, `fixed_scale`).
3. **Non-interactive madspace install** — route the installer's prompts through MG5's `cmd.ask` (falling back to `input()`), add `-f`/PYTHONPATH/DEVNULL propagation, and drive `generate_events` from the scripted card commands after `launch` so a command-file run no longer hangs on the installer.
4. **Find cmake from HEPTools** — before the source build, discover a cmake ≥3.15 (`MADGRAPH_CMAKE`/`CMAKE` → system → `heptools_install_dir` via env/MG5 config → default `<MG5>/HEPTools`) and put it on the build PATH.

## Testing

- `TestRunCardMG7` unit tests (12) + existing banner tests pass via `./tests/test_manager.py`.
- End-to-end `output mg7` for `p p > t t~` (hadronic defaults) and `e+ e- > mu+ mu-` (leptonic defaults, jet cuts dropped); produced `run_card.toml` is valid TOML and process-adjusted.
- Interactive `set`/units/shortcuts verified by driving the real `ask_edit_card_static` flow.
- Installer prompt force/pipe behaviour and cmake discovery (env var + `~/.mg5` config) verified.

## Notes

- Drafted for review; not yet exercised through a full first-time madspace source build.
- Parameters are addressed as `set <section>.<key>` because TOML keys are only unique within a section.


madgraph/iolibs/template_files/mg7/gridpack.py
```diff
@@ -8,9 +8,16 @@
...
-    # load run card and metadata
-    with open(os.path.join("Cards", "run_card.toml"), "rb") as f:
-        run_card = tomllib.load(f)
+    # load run card and metadata. Use the RunCardMG7 representation when the
+    # madgraph package is importable; gridpacks are meant to be portable, so
+    # fall back to a plain tomllib parse otherwise (the card is the same TOML).
+    run_card_path = os.path.join("Cards", "grid_run_card.toml")
+    try:
+        from madgraph.various.banner import RunCardMG7
+        run_card = RunCardMG7(run_card_path)
```
theoheimel, line 41: This never happens because madgraph is not added to the path. Do we want to keep it or just always use the tomllib one to be completely independent of the main MG code?

theoheimel:
I'm critical of the last two commits (1d23fe8 and 65ae188)
- accounting for the possibility of no ms.SOURCE_HASH adds a lot of complexity to the code, and we shouldn't be worried about backward-compatibility at the pre-alpha stage
- auto-compiling madspace if there are changes makes development quite annoying because it prevents us from running mg7 when there are unfinished changes that are not yet ready to compile. Maybe it makes sense for releases, but in development mode, this should stay just a warning

oliviermattelaer:
I have drop them ! Thanks

theoheimel:
Thanks! I had a look at the rest. I think it can be merged

oliviermattelaer:
Thanks I have added the auto-width functionality (relying on madevent for three-body decay).
And will likely want to check how to make those 5 red CI to a stage that they are either green or at least informative, before merging.

Note that this branch also has issue for tt~jj with top decay (added to the hall of fame)
 https://github.com/MadGraphTeam/MadGraph7/issues/13#issuecomment-4922778008

theoheimel:
The top decay error seems to be there in all branches, so that's an independent issue. I'll have a look at that

PR #26, opened 2026-07-04
Allow a save check_flavor at the helasdiagram level.

oliviermattelaer:
Note that this is much better to call such type of function at HelasMatrixElement level since they can reuse information (propagator) between different diagram rather than forcing recomputation


theoheimel:
Everything seems to work now. I didn't see any nonsensical flavors or missing flavor indices.

Please check my commit whether I am using the flavor checking API in the intended way.

PR #29, opened 2026-07-06
Use hashes to ensure madspace compatibility, options to bundle madspace in gridpacks

theoheimel:
Use hashes of madspace source files to ensure compatibilty
- of installed madspace and madspace source
- of madspace used to generated and run a gridpack

Options to include madspace in gridpacks:
- use system-wide installation
- include madspace build
- include madspace source (also in addition to build)

### DIFF
--- a/madspace/source_hash.py
+++ b/madspace/source_hash.py
@@ -1,0 +1,44 @@
+import hashlib
+from pathlib import Path
+
+_ROOT = Path(__file__).resolve().parent
+
+_SOURCE_DIRS = ("src", "include", "madspace")
+_SOURCE_FILES = ("CMakeLists.txt", "instruction_set.yaml", "pyproject.toml")
+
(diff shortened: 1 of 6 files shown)


madspace/source_hash.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+def compute_source_hash() -> str:
+    paths = sorted(_iter_source_files(), key=lambda p: p.relative_to(_ROOT).as_posix())
+    digest = hashlib.sha256()
+    for path in paths:
+        digest.update(path.relative_to(_ROOT).as_posix().encode("utf-8"))
+        digest.update(b"\0")
+        with open(path, "rb") as f:
+            digest.update(hashlib.file_digest(f, "sha256").digest())
```
oliviermattelaer, line 38: What is the minimum python version that we do target for madspace?
this line require python3.11 (ubuntu22 is not compatible already actually)

oliviermattelaer (approved):

PR #30, opened 2026-07-08
pre-compute contributing flavor

oliviermattelaer:
refactor the flavor such that we directly store at generate time the contributing flavor


theoheimel (approved):
I tested a few processes, cross sections and lists of active flavors seem to be correct.

PR #32, opened 2026-07-09
Add systematic tests to mg7

### LINKED ISSUE #28: Systematically testing MG7
To do:
- [ ] collect MG5 baseline results here
- [ ] add CI to check cross sections

Qubitol:
This implements #28.

Tests are defined in the metadata file `tests/check_xsec_processes/reference.json`.
All the processes listed in #28 are reported, but we may want to run a smaller set.

Additionally, the MadSpace installation has been made into a single shared action, that can be used by multiple workflows to build, install, and cache it, without code duplication.

**TODO** We need to decide when to run the workflow, if manually, on push, on push to dev/main, or similar.


oliviermattelaer:
Hi @Qubitol.

Would it be possible to keep this within the test_manager.py framework ?
The constraints are that:
1) all tests are within a class that depend of unitest.TestCase
2) each test/process is associated to a function of that class starting with test_ 
3) put the files in parralel_test (or acceptance_test)

In principle then the CI can then use test_manager.py to launch such test (which will be nice when I would implement a nice action for running such tests (which is not the case for the moment).
The main point is to have a single way to run all the tests and to be sure that all tests can be run locally.
(in top of the CI obviously)

After that how do you want to proceed?
Do you want to merge this as failing CI? and fix those in another branch? or let Theo fix those in this branch?

Qubitol:
I agree on all the points you raised, I will fix this to make sure it depends on the test manager.

Regarding the failing: if it is related only to the physics processes themselves, then my idea would be to merge as failing CI, and we can check each process separately until it satisfies the cross section.

theoheimel:
Maybe we should define the threshold based on the MC integration error, i.e. something like `(mean - mean_ref) / (sigma**2 + sigma_ref**2)**0.5 < thresh`

Qubitol:
Yes, that's a good idea.
Also, at this point, these tests should use the test manager.
Also, there are two points to discuss:
- do we allow generating different numbers of events for different processes? I was thinking maybe some need more events to better *converge* (even though the *weighted threshold* would make this requirement less stringent)
- when do we want to run these tests? At every push? Or only for commits to `main` or commits mentioning `validation` explicitly (it can be a trigger to run it now and then), or only manually?

oliviermattelaer:
At this stage I would be agressive on when to run those.
This means every push and every PR which are from external contributor (since they do not have the push CI in that case -- this is the reason why so many test are skipped actually)

If this is too much, then we will be able to reduce that in the future

theoheimel:
Regarding the first point: we should have some generic way to change run card options for some processes, like cuts for the ones with decays, or number of events for other ones should we find that is necessary. Could be as simple as a list of key-value pairs added to the json, and then the `set` command is applied for those.

oliviermattelaer:
I would propose to merge this, right away and then do a follow up PR where we handle the way to use the set command.
If you agree please merge this. (Thanks a lot @Qubitol )

Qubitol:
> I would propose to merge this, right away and then do a follow up PR where we handle the way to use the set command. If you agree please merge this. (Thanks a lot @Qubitol )

Ok, good for me, then.
Let's merge.

PR #34, opened 2026-07-13
Merging more integration channels within and across subprocesses

theoheimel:
To reduce MadNIS training times and trainable parameters, and make better use of its flexibility, allow for more aggressive merging of integration channels.

- [x] Within subprocesses: merge channels with different physics but identical phase-space topology. For instance, treat quark and gluon propagators the same at the phase-space level
- [x] Across subprocesses: allow merging channels with same topology, use flavor sampling to choose matrix element
  - [x] write out metadata to new json file
  - [x] add support for multiple matrix elements per channel in madspace
- [x] add GPU support
- [ ] add test for the jet tags


oliviermattelaer:
Should not we include the use of crossing symmetry in this list too?

theoheimel:
I'd say crossing symmetry is related to this, but independent enough to have its own PR. We should also consider to implement the flavor-dependent sign flipping and permuting on the matrix element side for crossing symmetry. This would make it easier to reuse for things like reweighting and standalone use. It would also make testing easier, because you can just run checks for individual phase-space points and compare it to the non-crossed implementation. Then for the integrator it would just look like you have less subprocesses with more flavors, and it does not have to know anything about crossing symmetry.

theoheimel:
This is working now and seems to give the correct cross sections, both in subprocesses-merged and classic mode. We should probably wait for #32 before we merge so we can test this more systematically.

@oliviermattelaer Can you have a look whether what I am doing with the `IdentifyJetTag` and `IdentifySGJetTag` looks reasonable to you?

oliviermattelaer:
I would say that from the name "Jet" you should exclude colorless particle.
since for the moment, this will also merge quark and lepton (I guess).

Please also add some unittest (and wire them within the CI) for such type of class.
Those are complicated rules and it is worth to have them checked explicitly and cross-checked/pin in future version.

theoheimel:
Yes, it will probably merge quarks and leptons. I'm not completely sure if that is a good thing or a bad thing, given that it would result in the same phase space topology. In case we want to keep that behavior, we should probably rename it so something like IdentifyTopologyTag.

Some tests will be helpful for sure, added it to the to do list.

PR #33, opened 2026-07-11
more agressive pruning of diagram without flavor

oliviermattelaer:
run the trimming
1) before color computation now
2) in all cases.
By default this also remove diagram only possible for symmetric flavor that the one considered.
the standalone --mask=False does not trim those last one for a simplify debugging/check

PR #31, opened 2026-07-08
Postprocessing linking

oliviermattelaer:
This include the run_card refactoring (that needs to be merged first).

But modifies the question such that we can decide to run Pythia8/MA5/Delphes/MadSpin/...
reweighting is still linked to the old module (need to be replaced by MadTrex).

One still need to check that all linking are working (but claude tells me that i can not write valid lhef file locally from madspace ...)

# Linkable tools status
- [x] Pythia8 
- [x] MadAnalysis5
- [x] time_of_flight
- [x] Delphes
- [x] Rivet/contur
- [x] systematics (miss mgrwgt information)
- [x] MadSpin (missing proc_card in banner) 
- [x] reweight

PR #42, opened 2026-07-15
Tmp merged

oliviermattelaer:


oliviermattelaer:
This is not designed to be merged per say.
This PR is the sum of #31 #32 #33 

It is just to see what is the current status with madspace CI when all of them are merged into a single branch.
This show that only 3 CI are not working anymore (3jets) and (ttx_decay)

PR #48, opened 2026-07-23
Add run_card seed for reproducible mg7 CPU event generation

oliviermattelaer:
@theoheimel, I know that you want to look at that yourself, so please reject the PR as soon as convenient. But here is the status that I do have so far:

## Summary

Adds a `seed` field to the mg7 run_card that makes **CPU event generation reproducible**: the same `(seed, thread-pool sizes)` reproduces **bit-identically**, and distinct seeds produce **statistically independent** results. A seed of `0` (the default) preserves the previous non-deterministic behaviour, so nothing changes for existing workflows.

### Why

Today `madspace` seeds every RNG from `std::random_device` with no seed input, and draws randomness from thread-local generators dispatched to the pool in non-deterministic order. As a result mg7 runs can never be reproduced — which blocks debugging, regression testing, and gridpack-style "same seed → same events" workflows.

## What changed

**Seed plumbing**
- `[run] seed` in the run_card + `RunCardMG7` default (`gridpack=True`).
- Per-context seed derivation (splitmix64) in `madevent.py`; a `--seed` flag in `gridpack.py`.
- `Context(seed=…)` exposed through the Python bindings; `seeded_rng()` / `mix_seed()` helpers in `util.hpp`; the driver-level `random_device` sites reseeded from the context seed.

**Determinism (gated on `seed != 0`, so the default path is untouched)**
- **Per-job RNG** keyed by logical job identity `(channel, sequence, phase)` via `Runtime::begin/end_job_rng`, so a job's randomness no longer depends on which worker thread runs it.
- **Deterministic commit ordering**: `survey`/`generate` harvest jobs in `job_id` order with unweighting folded in synchronously, making cross-section sums, max-weight snapshots, event order and stopping deterministic. Expensive matrix-element evaluation still runs fully in parallel; only the cheap commit step is ordered.

## Guarantee

| Setting | Result |
|---|---|
| Same `seed` + same thread-pool sizes | bit-identical output |
| Different `seed` | statistically independent output |
| `seed = 0` | non-deterministic (previous behaviour) |
| Different `cpu_thread_pool_size` | differs, but each is deterministic for its own thread count |

For `output_format = lhe`, bit-identical output additionally requires `combine_thread_pool_size = 1` (the parallel LHE *combine* is not yet deterministic). `compact_npy` and `lhe_npy` are reproducible at any combine size. This is documented in the run_card comment.

## Validation

End-to-end through the real mg7 driver (`bin/generate_events`) on `p p > t t~` with `dummy_matrix_element = true`, comparing `events.npy` byte-for-byte:

| Run | seed | threads | events | `events.npy` sha256 |
|---|---|---|---|---|
| A | 12345 | 4 | 500 | `93a535b4…` |
| B | 12345 | 4 | 500 | `93a535b4…` ✅ identical to A |
| C | 99999 | 4 | 500 | `697f09e1…` differs |
| D | 0 | 4 | 500 | `c4b0d5ac…` |
| E | 0 | 4 | 500 | `1cd4400a…` ✅ differs from D (default stays racy) |
| F | 777 | 8 | 3000 | `931a2ab4…` |
| G | 777 | 8 | 3000 | `931a2ab4…` ✅ identical to F |

Same seed reproduces bit-for-bit **at 4 and 8 threads**; different seeds differ; and `seed = 0` differs run-to-run — confirming the multi-threaded path is genuinely non-deterministic by default and that reproducibility comes from this change, not from a trivially serial workload.

Reproduce:
```bash
export LHAPDF_DATA_PATH=/path/to/LHAPDF        # dir containing the NNPDF set
# in Cards/run_card.toml: seed = 12345, dummy_matrix_element = true,
#   cpu_thread_pool_size = 4, output_format = compact_npy
bin/generate_events -f          # run twice
shasum -a256 Events/run_*/events.npy   # the two seed=12345 runs match
```

## Testing

New `madspace/tests/test_determinism.py` (8 tests) exercises the seeding through a real `random` op (`Context.seed → seeded_rng → op_random`): same seed reproducible, different seeds differ and are uncorrelated, adjacent seeds independent (guards the `std::seed_seq` mixing), uniform draws, and `seed = 0` non-deterministic.

> Note: the raw `FunctionRuntime` path is only guaranteed reproducible single-threaded, so the unit test asserts that; multi-thread *generation* determinism is covered by the end-to-end runs above rather than by this unit test.

## Scope / follow-ups

- **CPU only.** The GPU path is unchanged (still `random_device`) and out of scope here.
- **Fixed thread count.** Reproducibility holds for a given `cpu_thread_pool_size`; changing it changes the result (each deterministic).
- Not yet done (optional follow-ups): make the parallel `lhe` combine deterministic to drop the `combine_thread_pool_size = 1` caveat; add a portable, CI-runnable end-to-end determinism test (needs LHAPDF data + a process fixture).

### DIFF
--- a/madgraph/iolibs/template_files/mg7/gridpack.py
+++ b/madgraph/iolibs/template_files/mg7/gridpack.py
@@ -31,6 +31,24 @@ import json
 import tomllib
 import argparse
 
+
+def derive_seed(base_seed: int, index: int) -> int:
+    """Derive an independent per-context stream seed from the base seed.
+
+    Returns 0 (seed non-deterministically) when ``base_seed`` is 0. Otherwise the
+    base seed and context ``index`` are mixed with splitmix64 so distinct contexts
+    get well-separated streams and different base seeds never collide.
+    """
+    if base_seed == 0:
+        return 0
+    mask = (1 << 64) - 1
+    x = (base_seed + index * 0x9E3779B97F4A7C15) & mask
+    x = ((x ^ (x >> 30)) * 0xBF58476D1CE4E5B9) & mask
+    x = ((x ^ (x >> 27)) * 0x94D049BB133111EB) & mask
+    x = x ^ (x >> 31)
+    return x or 1
+
+
 def main() -> None:
     # load run card and metadata. Use the RunCardMG7 representation when the
     # madgraph package is importable; gridpacks are meant to be portable, so
@@ -59,6 +77,11 @@ def main() -> None:
     # parse command line arguments
     parser = argparse.ArgumentParser()
     parser.add_argument("--run_name", type=str, default=run_args["run_name"])
+    parser.add_argument(
+        "--seed", type=int, default=run_args.get("seed", 0),
+        help="0 draws a fresh random seed; a positive integer makes the run "
+             "reproducible (bit-identical output requires --cpu_thread_pool_size 1)"
+    )
     parser.add_argument("--device", type=str, nargs="*")
     parser.add_argument(
         "--cpu_thread_pool_size", type=int, default=run_args["cpu_thread_pool_size"]
--- a/madgraph/iolibs/template_files/mg7/gridpack.py
+++ b/madgraph/iolibs/template_files/mg7/gridpack.py
@@ -88,7 +88,8 @@ def main() -> None:
     parser.add_argument(
         "--seed", type=int, default=run_args.get("seed", 0),
         help="0 draws a fresh random seed; a positive integer makes the run "
-             "reproducible (bit-identical output requires --cpu_thread_pool_size 1)"
+             "reproducible at any --cpu_thread_pool_size (for --output_format lhe "
+             "also set combine_thread_pool_size = 1)"
     )
     parser.add_argument("--device", type=str, nargs="*")
     parser.add_argument(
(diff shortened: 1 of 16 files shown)


theoheimel:
I think the key ideas are correct, i.e. having a derived per-batch seed and ensuring a consistent commit order. However in its current form this will perform quite poorly for inclusive jobs where evaluation times differ a lot between channels. I have a few ideas how this can be fixed elegantly, but need to think about it some more. Ideally it should be fast enough so that we can always use the reproducible variant.

oliviermattelaer:
In any case experimentalist will always use reproducible mode. So this should be the default (whatever we think about it, this is not negotiable for them).

So I played with a per channel scheduler that lift that locking and tested on top quark pair (up to two additional jet)
Scheduler | wall | cpu | efficiency | integral | samples
-- | -- | -- | -- | -- | --
OLD (global-order) | 3.48s | 29.15s | 8.38× / 8 | 1.15037e9 | 47.9M
NEW (per-channel) | 3.46s | 28.90s | 8.35× / 8 | 1.15035e9 | 48.0M

With 224 channels and millions of events there's an enormous amount of independent work in flight at all times. When the global commit cursor lands on an expensive tt̄jj job, the 16-deep in-flight buffer plus the sheer number of other ready channels means threads always have something to run — the cursor stall is negligible against the total work. The cost skew here is "most work is expensive" (the 190 tt̄jj channels dominate), which parallelizes fine; it's not "cheap events stuck behind rare expensive ones."

I can push it if you want.

theoheimel:
Yes, please push it, then I can also play around with it. I think we have to test it for a bunch of setups, especially CPUs with a lot of cores in combination with only few channels. This is where I would expect a per-channel scheduling to perform better

PR #49, opened 2026-07-23
Crossing symmetry

oliviermattelaer:
# Implementation:

- [x] New default (--no_crossing -> --use_crossing)
- [x] New default value (use_crossing is ON by default 
    - [ ] @theoheimel we might want to backtrack that if you are not ready for madspace
- [x]  encoding choice via larger range for FLAV_IDX
     -  FLAV_IDX = cross*nflav + flav where flav is the previous flavor index to use
     -  cross = flip1*(nexternal+1)+flip2, where flip1 is the position (1 encoding) of the final state particle that is flip with particle 1, and flip2 is the position of the final state particle flip with particle2 (0 means no flipping)
        -  [ ] encoding fail to capture all possible crossing (see dedicated post below).
- [x] change NHEL that instead of a lookup table (complicated for crossing) is a radix encoding of the helicity 
- [x] change color index such that it also correspond to an encoding 
- [x] supported in standalone_fortran
    - [x] check_sa.f does contains a demo code (not active) on how to use such feature
    - [x] f2py compatible getting the index from the PDG requested
    - [x] can be check via the command "check crossing p p > j j j" 
    - [x] C symmetry of helicity are now checked 
- [x] supported in standalone_cpp
   - [x] check_sa.cc does contains a demo code (not active) on how to use such feature
   - [x] can be check via the command "check crossing p p > j j j --output=standalone_cpp"
   - [x] C symmetry of helicity are now checked 
- [x] supported in standalone_mg7
   - [x] can be check via the command "check crossing p p > j j j --output=standalone_mg7 --simd=sse4 --precision=d"
   - [x] check command works for other simd format
   - [x] change SIMD strategy (seed post below) 
   - [x] C symmetry of helicity are now checked 
- [x] moves the T-channel propagator width treatment into ALOHA. 
    - Outside the complex-mass scheme, the ALOHA propagator routine now drops the width i·M·Γ for spacelike (P²<0) momenta at runtime (correct t-channel treatment), replacing the old codegen call-rewrite (which only affected madevent's fk_ widths — standalone/cpp never got it)
    - [x] new mg option: zerowidth_tchannel  to control if this need to be activated (old madevent trigger is crashing now!)
- [x] New width treatment zerowidth_external (default True): set the width to zero (internally to the matrix-element) for any particle that appears in the final state)
   - [x] not applied if a particle is marked as offshell (*, syntax supported for standalone)  
- [x] Madevent handling:
    - [x] Crossing within the same group: 
        - a single matrix.f
        - multiple auto_dsigX.f
        - metadata generated as in the case of grouping
     - [x] crossing-equivalent MEs in different P directories share one compiled ME via symlink; 
         - the dependent keeps its own PDF/PS/leshouche/configs and routes its auto_dsig to the base SMATRIX.
         - Per-flavour event maps: flavour route (DSIG_XGROUTE), CONFIGMAP (multichannel diagram pairing), COLMAP, 
helmap.
         - Shared base .o via crossgroup.mk + a single parallel top-level makefile for all P dirs.
         - hadronic p p also uses cross-group sharing; crossing limitation tag extended to cross-group. 
     - [x] Helicity recycling:  a crossing base bakes the shared optim over `G_base U tau(G_base)` (`crossgroup_helunion.dat`); wavefunction recycling is retained.
        - helicity needs a flipping associated to the permutation 
        - `tau` is that flipping applied *in place* (the crossing's NSF sign flips), and it is the ONLY part of the transform the recycled optim can realise: it bakes its helicity configs into the HELAS calls and receives just `(PUSE, IC)` -- IC carries the signs, nothing carries the slot permutation.
        - the permuted map (the GHREMAP `sigma` that `matrix<b>_orig.f` applies via `APPLY_CROSSING_TABLE`) is the wrong one here. Baking `G_base U sigma(G_base)` cost -28.5% on `q q~ > q q~`; fixed in db46ee28c, pinned run-free by `TestCrossingRecycledHelicityUnion`.
        - AMP2 (multi-channel) and JAMP2 (colour flow) -- not the |M|^2 sum -- are gated on `TS(K).NE.0D0` for a crossing base, so each member weights channels and colour over exactly its own good set. A shared optim necessarily holds rows that are dead for whichever member is calling.
      - [x] C symmetry of helicity are now checked compatible with helicity recycling
         - [x] checked that event helicity are correctly written
      - [x] Block beam polarisation + EVA under crossing (clear crash rather than silent wrong results).
      - [x] handled only for the production side in case of decay chain syntax. 
- [x] New tests/acceptance_tests/test_standalone_cross_symmetry.py (multi-process crossing checks) and additions to test_standalone_madevent_consistency.py.
    
# Validation:
* p p > j j bit-identical; p p > j j j matches; colour conserved.
* p p > t t~ j j: completes (was crashing), total within MC error, colour 0/10000 unphysical.
* p p > w+ w- j j: 22.74 vs 22.91 pb (<1σ), colour 0/2000 unphysical; QCD=0: 0.7315 vs 0.7326 pb.
* Lepton/photon EP EM > EP EM: now matches reference (0.2675, was ~1.8% low).
* q q~ > q q~ (q=u d s c, apply_flavor_grouping False, group_subprocesses True, MLM ickkw=1/xqcut=20): 5188690 +- 15750 vs the 3.7.2 reference 5184588 +- 2971 -- `acceptancetest_flavor::test_flavor_grouping_consistency_mlm` green on all four settings (the crossing+recycling one was 3706990 +- 10989, 105.8 sigma). Colour flow held: `TestMadeventCrossingBaseColorFlow` and `TestMadeventRouterColorSelection` pass, `g g > u u~` crossing-on and `--use_crossing=False` both 3.023e6 +- 1406 pb.
* In progress: check the rate  in the lhef of 
    - [x] check the rate  in the lhef of  flavor
    - [x] icheck the rate  in the lhef of nitial state
    - [x] check the rate  in the lhef of  color
    - [x] check the rate  in the lhef of helicity (CURRENTLY FAILING --under investigation--)
    - [x] check density matrix computation 
       
# Known limitations / follow-ups
 - [x] IOTest references for the changed ALOHA propagator routines (TestAlohaWriter/*) still need regenerating.
 - [ ] IOTest for madevent/standalone
 - [x] test_standalone_madevent_consistency.py uses group_subprocesses False, which the non-group madevent exporter doesn't support under the default use_crossing=True — needs --use_crossing=False there (or non-group crossing support).
 - [x] need a full run of the test suite (some reweighting one fail for example)
# merge option
  - merge now but set back the default to no_crossing
  - implement in madspace before merging


theoheimel:
Do we need any new functionality for that on the madspace side or will it just affect the generation of the subprocesses.json file and we will have a few more flavors within each subprocess?

oliviermattelaer:
The things that was needed for madevent were multiple map to correctly connect the information
1) a map over the helicity (madevent use an helicity index which does not map trivially when crossing are happening -- important for the writing of the lhef)
2) a map over the color (also for the writing of the lhef)
3) a map for the sde (when the same matrix-element is used between grouping mode). This one I do not think that you will need (it is already contained in your json file)

So yes only subprocesses.json need to be modified. I forwarded your question to claude:

Short answer: It's not just "a few more flavors within each subprocess" — a crossing changes the initial state, so a crossed process can't live inside an existing subprocess's flavor list. But the extra work on the madspace side is small and localized: for the cross-section it's essentially just "pass the extended flavor id to the shared ME"; the phase-space machinery is untouched. The only genuinely new piece is for event generation (the per-crossing helicity/colour/channel index maps), and that's a design choice about where those tables live.

Why it isn't just more flavors in a subprocess. In subprocesses.json each subprocess has a fixed incoming/outgoing, and its flavors are what the exporter calls all_flavors_same_initial — different PDG assignments that share the same initial-state slots, hence the same kinematics/phase space. A crossing by definition moves a leg between the initial and final state, so a crossed process has a different incoming/outgoing (different PDF legs, generally different masses). It violates the same-initial invariant, so it must be its own subprocess entry, not an extra entry in the base's flavors list.

What actually changes in subprocesses.json. You get additional subprocess entries (the crossed ones), each with its own incoming/outgoing/channels/helicities and its own phase space, but pointing at the same me_path (shared compiled ME) plus one new field: the crossing / extended flavor id (cross*nflav + flav). This mirrors the Fortran madevent design exactly — each crossed process keeps its own phase-space setup and only the matrix-element object is shared.

What madspace needs — split by use case:

Cross-section / integration: essentially nothing new. Each crossed subprocess is an ordinary subprocess with its own incoming/outgoing, so phase space, PDFs, cuts, and channel sampling all work as today. The one requirement is that the ME call forwards the extended flavor id so the ME applies its internal crossing. The madmatrix/cudacpp backend already supports this (supports_crossing=True, extended flavor id) and returns the correct crossed |M|² internally — so no new PS math and no correctness work for the xsec.

Event generation (spin/colour in the event record): here there is something. The ME returns the selected helicity, colour flow, and multichannel channel in the base process's enumeration, but the event for the crossed process needs them in the crossed process's own convention. That's three small per-crossing maps — the direct analogues of the Fortran DSIG_XGHEL (helicity), COLMAP (colour), and CONFIGMAP (channel↔diagram). Decision to make: either emit them into subprocesses.json per crossed subprocess and have madspace apply them after the ME call, or bake them into the C++ ME. (Note CONFIGMAP is only an importance-sampling/variance matter, not correctness.)
Net: more subprocess entries (not more flavors per subprocess), each carrying a crossing id and sharing a me_path; zero new phase-space functionality; a trivial ME-call change to forward the extended flavor id for the xsec; and a small, well-defined set of index maps if/when you want correct spin+colour in the generated events.

Confirmed: the madmatrix/cudacpp ME currently returns the selected helicity, colour, and channel/iconfig in the base process's numbering — only the |M|² is crossed. So the crossed-process index maps (helicity/colour/channel, = Fortran DSIG_XGHEL/COLMAP/CONFIGMAP) are not yet handled by the ME. For cross-sections that's irrelevant; for event generation someone has to apply them — cleanest inside the C++ ME (the colour one especially, since it needs the base ME to publish its per-flow jamp2 and re-select in the dependent's colour basis), otherwise emitted into subprocesses.json and applied by madspace.

oliviermattelaer:
I'm looking to a new way to encode helicity to avoid to have to do a mapping. If it works we can try similar (but more complex?) idea for the color as well.

oliviermattelaer:
Done with having coding of NHEL representing the value of the helicity (rather than an arbitrary look-up table).
This is obviously slightly slower (need to encode/decode) but at the end much more practical (in particular with crossing but also for density matrix/...
Tell me what you think about such idea, I can think of implementing something similar for the color-flow.
(likely more complicated since this is how particles connect but likely doable)

# Documentation on the encoding to interface that with madspace:

Writing event helicities with the canonical helicity encoding (madspace)
This PR changes how a matrix element (ME) labels the helicity it selects for an event. A consumer that writes per-particle helicities (LHE SPINUP, the density matrix, …) must interpret that label with the rules below. Momenta and |M|² are unchanged — only the helicity label.

- The convention: the selected helicity is a self-describing canonical code
The ME no longer returns an arbitrary row index. It returns one 1-based integer — the canonical mixed-radix code — that fully describes the helicity of every external leg.

Per external leg k (1..nexternal):

NHSTATE[k] = number of physical helicity states (2 for a fermion / massless vector, 3 for a massive vector, 4/5 for spin-3/2, spin-2).
STATES[d][k] = leg k's helicity values, i.e. get_helicity_states(allow_reverse=True) for the Fortran backends (allow_reverse=False for C++/cudacpp — each backend is internally self-consistent, so decode with the same backend's tables).
Last external leg = least-significant digit (matches itertools.product / the old NHEL order):

code = 1 + Σ_k  digit[k] · Π_{j>k} NHSTATE[j]

- Decoding a code → per-leg helicities
w = Π_{j>ipart} NHSTATE[j]
digit = ((code - 1) // w) mod NHSTATE[ipart]
helicity[ipart] = STATES[digit + 1][ipart]
You don't have to reimplement it — every ME exposes it:

Fortran madevent (auto_dsig<i>.f): GET_NHEL<i>(code, ipart) → code>0: the helicity of leg ipart; code=0: NHSTATE[ipart].
Fortran standalone: DECODE_HEL(code, config), GET_NHEL(iden, table), SMATRIXHEL(P, code, flavor_idx, ans).
So the writer is just SPINUP[ipart] = get_nhel(code, ipart) for ipart = 1..nexternal.

- Without crossing
Nothing special: each subprocess ME returns a code in its own enumeration; decode as in §2.

- With crossing
A crossed subprocess shares a base ME, called with an extended index FLAV_IDX = cross·nflav + flav.
   - Option 1: Recommended: use the subprocess's own SMATRIX<i> / get_nhel<i> (crossing is transparent). The Fortran madevent relabels internally — the router / auto_dsig<i> permutes the base code's digits by the crossing permutation, so the code handed back is already in this subprocess's enumeration. Decode it with this subprocess's get_nhel<i>. Identical to §3, no crossing code on your side.

   - Option 2:  If you call the base ME directly (e.g. today's C++/cudacpp Bridge, §5), the code is in the base leg order. Convert with the permutation:
CALL GET_CROSS_PERM(FLAV_IDX, PERM, SGN, FLAV_USE)   ! PERM[k] = base leg landing in this subprocess's slot k (1-based)
for ipart in 1..nexternal:
    SPINUP[ipart] = get_nhel_base(code, PERM[ipart])  ! base ME's get_nhel, read at the permuted leg
Critical: copy the helicity value unchanged — do not multiply by SGN. SGN is the NSF/momentum sign; it belongs to |M|² (nh = nhel·nsf), not the label. Applying it double-counts and silently reverses every crossed leg's helicity — invisible on non-chiral final states, wrong for e.g. w+ w- j j. (dep_states[k] == base_states[PERM[k]], no reversal — the incoming→antiparticle wf flip and the crossing conjugate+swap cancel.)

- Backend status
Fortran standalone + f2py + madevent — fully in this PR; §4a works out of the box.
C++/cudacpp (standalone_mg7) — non-crossing already returns the canonical code (the selected ihel is the mixed-radix code). The crossed relabel is not yet wired into the C++ allselhel selection (a scoped follow-up), so for a crossed flavor apply §4b with the generated cross_perm / GET_CROSS_PERM until it lands.

- Regression guard
test_standalone_cross_symmetry.py::TestMadeventCrossingHelicity.test_w_helicity_asymmetry_ppwj — a small p p > w+ j run must write a physical, chirally-asymmetric W⁺ polarisation; a scrambled relabel breaks it.

# Summary:
tl;dr for madspace: the selected helicity is now a self-describing 1-based mixed-radix code; write SPINUP[ipart] = get_nhel(code, ipart). If you go through the per-subprocess SMATRIX<i>/get_nhel<i>, crossing is transparent. If you call the base ME directly (C++ today), read the base code at GET_CROSS_PERM's permuted legs — and never apply SGN to the label.

theoheimel:
Yes, I think using a smart encoding makes a lot of sense. If the encoding/decoding speed is an issue (I'd be surprised though), we can also consider a less compact but easier to decode representation with a fixed width of 3bits per particle for up to 8 helicity states.

oliviermattelaer:
Ok we agree, let me try to do the same for the color of the events then.

oliviermattelaer:
Done for the color encoding:

The matrix element emit a self-describing colour code (the colour analogue of the helicity encoding). The encoder is done on the MG7 side; the runtime that returns and decodes it is the madspace/cudacpp half described below.

# What the ME now provides
- subprocesses.json (per subprocess): color_slots = {"color":[...], "acolor":[...]} — colour/anticolour slot leg numbers (1-based), flow-independent; color_codes — code of each flow (informational); color_flows — old ICOLUP tags, kept as fallback
- coloramps.h (per ME) : constexpr int colorflowcode[ncolor] + constexpr bool colorflowcode_valid

The code of the selected flow is colorflowcode[icolC], where icolC is the 0-based flow index the ME already picks in select_col_and_diag (allselcol = icolC + 1). Same integer on all three backends — the fortran colorflow.inc, the JSON color_codes, and the C++ colorflowcode[] agree byte-for-byte (u u~ > g g → {7, 11}).

# The encoding

Order colour slots and anticolour slots by leg on the initial-state-flipped flow (every colour connects to an anticolour). Then code = Σᵢ digitᵢ·Nⁱ (N = #anticolour slots), where digitᵢ is the anticolour-slot index that colour slot i connects to.

# Two runtime pieces to implement (specified, not yet built):
- api.cpp — return the code. It currently hardcodes color_out = 0 (mg7 colour is a stub today — the same gap as helicity_out = 0). Once the event's colour is selected: (likely fixed in the latest main)

- madspace LHECompleter — decode instead of table-lookup. Colour enters complete_event_data in exactly one place: the per-external-leg (colour, anticolour) list of the selected flow (color_flows[flavor][color_index][perm_index]). Replace that with a decode of the returned code producing the same length-nexternal array in matrix leg order; everything downstream (internal-line reconstruction, cancellation) is unchanged.
```
for (int i = 0; i < N; ++i) {                 // N = acolslots.size()
    int digit = (code / ipow(N,i)) % N;
    int tag = 501 + i;                        // fresh; only which legs SHARE a tag matters
    int cleg = colslots[i], aleg = acolslots[digit];
    // colour of cleg / anticolour of aleg, but SWAP those positions on initial-state legs
    (cleg <= n_initial ? anticol(cleg) : col(cleg)) = tag;
    (aleg <= n_initial ? col(aleg) : anticol(aleg)) = tag;
}
```

# Three things to know
- The initial-state position swap is the one convention to get exactly right — it undoes the flip that defines the code, and it matches the fortran addmothers decoder.
- Tag numbers are free. Downstream relabels and cancels, so 501+i per event is equivalent to the old color_flows numbering. Events are colour-equivalent, not byte-identical (same as the fortran side).
- Fallback: colorflowcode_valid == false for colour sextets and epsilon structures (color_codes/color_slots null). Keep the color_flows lookup there. No SM tree process hits this.

# color_flows — can be removed (but need validation/developement for sextet/epsilon_ijk, so fallback to it in those case)
Redundant for every valid-code subprocess (decode gives the externals; internal lines derive from them), but required as the sextet/epsilon fallback, so it stays in the JSON.

# Validation boundary
Encoder validated (JSON + coloramps.h cross-checked against the fortran output; CPPProcess.cc compiles). The api.cpp/madspace pieces above need a cudacpp build + an mg7 event run to validate end-to-end — the same path the helicity encode is deferred behind.

oliviermattelaer:
@Qubitol, I add to take some decision concerning the nice handling of helicity/crossing to be optimised with SIMD, here is the documentation of such strategy. Would be nice if you can comment on this.


# Per‑lane helicity: the SIMD strategy for `standalone_mg7` crossing

*Backend: `madmatrix` (cudacpp). Commit `4663f1855` on `claude/fortran-cross-symmetry-3f13f3`.*

## 1. Background — crossing + the good‑helicity filter

Crossing symmetry lets one **base** matrix element serve all of its crossed
subprocesses. `g q → t t̄ q` already computes `g q̄ → t t̄ q̄`, `q q̄ → t t̄ g`, … —
same diagrams, relabelled legs. In `standalone_mg7` this is carried by an
**extended flavor id** per event:

```
flavorID = cross * nmaxflavor + flav        // cross ∈ [0, (npar+1)²)
```

`cross = 0` is the identity (base process); the momenta of a crossed event are
permuted into the base slot order per event (`_crossing_preamble`, *deliberately
not vectorised* — see §2), and the swapped legs' NSF sign is flipped.

Orthogonally, cudacpp uses a **good‑helicity filter**: `sigmaKin_getGoodHel`
samples a few events and keeps only the helicity rows that contribute; `sigmaKin`
then loops over that reduced set. The tension this note is about is *how the
good‑helicity filter interacts with crossing in a SIMD/GPU execution model*.

## 2. Why the original design used a *union*

cudacpp processes events in SIMD lockstep. `sigmaKin` loops **once** over a single
shared list `cGoodHel`, and for each helicity `calculate_jamps` computes **all**
events in the vector:

```cpp
for (int ighel = 0; ighel < cNGoodHel; ighel++) {
  const int ihel = cGoodHel[ighel];              // ONE scalar helicity, all lanes
  calculate_jamps(ihel, …, iflavorVec, …);       // whole SIMD vector
}
```

The catch: **events in one SIMD page share `flav_use` but may carry different
crossings.** (That is exactly why `_crossing_preamble` gathers each lane's
permuted momenta in a scalar, per‑event loop.) A *per‑crossing* good‑helicity
gate therefore cannot be applied at this shared loop level — different lanes
would want different helicity subsets.

The original solution was the **union**: `getGoodHel` scans every
`ncross * nflav` combination and ORs the result, so `cGoodHel` is the union of
good rows over all crossings. Rows that vanish for a given lane's crossing simply
contribute `0`. Correct, and SIMD‑clean — but a superset.

## 3. The cost of the union

The per‑crossing good sets are *permutations of each other* (crossing is a
relabelling), so they all have the **same size** but sit on **different rows** of
the table. On a chiral/EW process the good set is small yet the crossings scatter
it across the whole table, so the union blows up.

Measured, `p p > w+ j`, the `gQ` directory (`ncomb = 24`):

| | helicity rows computed / event |
|---|---|
| any single crossing (base good set) | **6** |
| union over all 25 crossings | **24** (all of them) |

→ the union computes **4× more helicities per event than any single crossing
needs**. (For pure‑QCD processes the base set is already near‑full, so the union
is barely larger and the win is small — the payoff is on EW/chiral matrix
elements.)

## 4. The strategy: per‑lane helicity

The key observation: **helicity only ever enters the external wavefunctions**
(`ixxxxx` / `oxxxxx` / `vxxxxx`). Everything after them — internal wavefunctions,
`FFV*` vertices, colour decomposition (`get_amp`) — is **helicity‑agnostic and
already fully SIMD**.

So we do *not* need a shared helicity across the vector. Instead:

* loop once over the **per‑crossing good‑hel count** (not the union);
* at iteration `ighel`, each lane uses **its own** crossing's `ighel`‑th good
  helicity — i.e. the helicity becomes a **per‑lane** quantity;
* the external wavefunction for a leg is built for each of the leg's helicity
  states and **mask‑blended** lane‑wise (does this lane want state `v`?);
* `get_amp` then runs full‑width on the blended wavefunctions — **no lane masking
  downstream, no union waste.**

This is the SIMD‑native realisation of the same *shared‑filter* idea the Fortran
and `standalone_cpp` backends express with a per‑event gate (`CROSS_GHIDX` /
`ghfilt`): each event ends up evaluating exactly its own good helicities.

### External block, per leg (schematic)

```cpp
{ static const int _st[NS] = { <helicity states of this leg> };   // e.g. {-1,1} or {-1,0,1}
  for (int _vi = 0; _vi < NS; _vi++) {
    const int _v = _st[_vi];
    ROUTINE(xmom, MASS, _v, +nsf, …, aloha_x[0], s);              // build wf at helicity _v, both NSF
    ROUTINE(xmom, MASS, _v, -nsf, …, aloha_x[1], s);
    const fptype_sv _sp = (icsign[s] + 1)*0.5, _sm = (1 - icsign[s])*0.5;   // per-lane NSF blend
    fptype_sv _hm{};
    for (int _ie = 0; _ie < neppV; _ie++) {                       // per-lane helicity mask
      int _hr = (_ighel < 0) ? ihel
              : cGoodHelOfCross[ iflavorVec[ievt0 + _ie] / nmaxflavor ][_ighel];   // (padded to -1)
      ((fptype*)&_hm)[_ie] = (_hr >= 0 && cHel[_hr][s] == _v) ? 1. : 0.;
    }
    // accumulate the lane-selected wavefunction (get_amp is unchanged after this)
    w_sv[s]    += _hm * (_sp*w_x[0]    + _sm*w_x[1]);
    pvec_sv[s] += _hm * (_sp*pvec_x[0] + _sm*pvec_x[1]);
  } }
```

Scalars (spin 1) carry no helicity, so their block stays the plain NSF blend.
The per‑leg state list comes from `_hel_state_values(spin, mass)` (fermion →
`{-1,1}`, massless vector → `{-1,1}`, massive vector → `{-1,0,1}`, spin‑2 too).

### Data structures

`getGoodHel` builds a **per‑cross** table during its existing scan:

```cpp
static int cGoodHelOfCross[ncross][ncomb];   // good rows per crossing
static int cNGoodPerCross[ncross];           // count per crossing
static int cNGoodMaxCross;                   // loop bound (= base count)
```

`sigmaKin`'s summation loop then runs `ighel < cNGoodMaxCross`.

## 5. The mixed‑precision subtlety (please look here, reviewers)

`calculate_jamps` processes `nParity` pages in a loop, and the external block runs
**once per page**:

```cpp
for (int iParity = 0; iParity < nParity; ++iParity) {   // nParity = 2 in mixed precision
  const int ievt0 = ievt00 + iParity * neppV;           // this page's first event
  … external block …
}
```

So the per‑lane helicity **must be derived per page** from `ievt0`. A first
implementation precomputed a `_ihlane[neppV]` array in `sigmaKin` for page 0 only;
in mixed precision (`nParity = 2`) that **silently zeroed the second page** —
every second neppV‑page of events came out `0` under divergent crossings.

The fix (and the current design): pass the good‑hel **index** `_ighel` to
`calculate_jamps`, and let the external block resolve `_hr` from
`iflavorVec[ievt0 + _ie]` **per page**. `_ighel < 0` falls back to the scalar
`ihel` (used by `getGoodHel`'s scan and by the non‑crossing path).

## 6. Scope and gating

* **C++ path only** — scalar‑SIMD *and* mixed precision. **GPU
  (`MGONGPUCPP_GPUIMPL`) stays on the union**; it isn't ported and needs a CUDA
  box to validate. (The CUDA async helicity loop is deliberately left on
  `cNGoodHel`; note it is textually identical to the C++ loop and is
  distinguished by the following line, `fptype* hAllJamps` vs `cxtype_sv jamp_sv`.)
* **Fully gated on `use_crossing`.** A non‑crossing build is **byte‑identical**:
  the OFF fills keep `cNGoodHel` / `cGoodHel[ighel]` / the scalar `ihel`, and the
  `_ighel` parameter is not even added to `calculate_jamps`.
* Codegen lives in `madmatrix/model_handling.py`
  (`get_madmatrix_crossing_dict` holes + `_crossing_external_block`) and two
  templates (`process_function_definitions.inc`, `process_sigmaKin_function.inc`).

## 7. Validation

`check_sa -f <flavorID>` forces **one** crossing for the whole run, so all lanes
share a crossing — it cannot exercise the very thing this change is about.
A small **mixed‑crossing driver** was used: patch the host `flvVec` to cycle
crossings per event (`MG_FLVMIX="c0,c1,…"`), so neighbouring lanes in a SIMD
vector carry *different* crossings.

Results on `cppsse4` (neppV = 2, via `sse2neon` on arm64):

* per‑event |M|² is **byte‑identical to the union build**, in **both**
  `FPTYPE=d` (pure double) and `FPTYPE=m` (mixed precision, `nParity=2`), with
  divergent crossings inside a SIMD vector;
* `check crossing --exporter=standalone_mg7 --simd=sse4`: **3/3** (`p p > w+ j`)
  and **4/4** (`p p > t t~ j`);
* `--use_crossing=False` emits zero per‑lane machinery (historical output).

Work per event drops from 24 → 6 helicity iterations on `w+ j` (~4× on the
dominant `get_amp`); even the external‑wavefunction call count nets down
(≈ 192 → 108) despite the per‑state blend, because the iteration count falls
faster than the per‑iteration cost rises.

## 8. Open items

* **GPU** per‑lane port (needs CUDA to validate).
* The reported per‑event helicity `allselhel` for *crossed* events (separate
  commit `1eb6ec73c`) is compile‑checked but runtime‑unvalidated — it does **not**
  affect |M|², only the LHE helicity label; it needs an event‑level cudacpp run
  against the Fortran backend to settle the permute‑vs‑NSF convention.

oliviermattelaer:
One subtelty to keep in mind (maybe we should change the crossing encoding?):
# Why `q q~ > t t~ q q~` keeps its own matrix element (and is *not* routed as a crossing of `q q > t t~ q q`)

**Context.** In `p p > t t~ j j` the group `P*_qq_ttxqq` contains two compiled
matrix elements, `matrix1_orig.f` (`Q Q > t t~ Q Q`) and `matrix2_orig.f`
(`Q Qx > t t~ Q Qx`), plus one crossing router (`Q~ Q~ > t t~ Q~ Q~`). A natural
question is why `Q Qx` is a *second* matrix element rather than a crossing of
`Q Q`, since `q q > t t~ q q` and `q q~ > t t~ q q~` are related by crossing.

**Short answer.** It is **not** a filtered/"no-flavour" diagram, and the
processes **are** genuine crossings. The one flavour that cannot be routed is the
flavour-changing annihilation `d d~ > t t~ u u~`, and the obstruction is a
**limitation of the crossing encoding** (it can only swap the two *initial* legs
with final ones — it has no final↔final permutation), not a physics or
diagram-content difference.

## The facts (verified)

Diagram counts match, so no diagram is dropped — these are true crossings:

| process              | `NGRAPHS` | crossing of …                       |
|----------------------|:---------:|-------------------------------------|
| `u u  > t t~ u u`    | 14        | —                                   |
| `u u~ > t t~ u u~`   | 14        | `u u > t t~ u u`  ✅                 |
| `d u  > t t~ d u`    | 7         | —                                   |
| `d d~ > t t~ u u~`   | 7         | `d u > t t~ d u`  ✅                 |

Reachability of the crossed signatures (from `compute_crossing_pdg_entries`):

```
u u~ > t t~ u u~   reachable from u u > t t~ u u   : True   (both leg orders)
d d~ > t t~ u u~   reachable from d u > t t~ d u   :
      (1,-1,6,-6, -2,  2)  =  d d~ > t t~ ū u      : True
      (1,-1,6,-6,  2, -2)  =  d d~ > t t~ u ū      : False   <-- the order the merged ME uses
```

## Why the merged `Q Qx` ME still cannot route

`Q Qx > t t~ Q Qx` bundles (via flavour grouping) three flavour classes. Two of
them **do** cross to `Q Q`:

* `d d~ > t t~ d d~`  (same flavour) → crossing of `d d > t t~ d d`  ✅
* `d u~ > t t~ d u~`  (mixed)        → crossing of `d u > t t~ d u`  ✅

The third is the flavour-changing annihilation `d d~ > t t~ u u~` (initial `d d̄`,
final `u ū`, via `d d̄ → g → t t̄` and `g → u ū`). It is a crossing of
`d u > t t~ d u`, but only with the two final same-flavour-group legs in the
order `(ū, u)`:

```
crossing of d u > t t~ d u delivers :  … , ū(5) , u(6)     -> sig (1,-1,6,-6,-2, 2)
merged Q Qx > t t~ Q Qx lists       :  … ,  u(5) , ū(6)    -> sig (1,-1,6,-6, 2,-2)   (Q at 5, Qx at 6)
```

Going from `(ū, u)` to `(u, ū)` is a **swap of the two final legs**. The crossing
code encodes a crossing as

```
CROSS = I*(NEXTERNAL+1) + J        # swap initial-1 with final-I, initial-2 with final-J
```

i.e. only crossings of the **two initial legs** with final ones — there is no
final↔final permutation in the encoding. Since `u` and `ū` are different
particles, feeding the base's crossing SMATRIX the dependent's momenta in the
wrong order would evaluate the wrong `|M|²`, so `partition_crossing_classes`
correctly refuses the match. Routing a merged member requires **every** flavour
to cross to a single base; this one flavour cannot, so the whole `Q Qx` ME stays
a base and gets its own `matrix<i>_orig.f` (and its own helicity-recycling run).

## Consequences / how it could be enabled

* The current output is **correct** — `Q Qx` genuinely needs a matrix element it
  can evaluate; nothing is double counted or lost.
* This is *not* the cross-group Track-B case: `g g > t t~ q q~` ↔ `q q~ > t t~ g g`
  *are* reused (symlinked base matrix element), because there the crossing is a
  plain initial↔final swap the encoding can express.
* To share `Q Q` with `Q Qx` one would have to **generalise the crossing** to also
  permute final legs (or have flavour grouping assign, for that flavour, the leg
  order the initial-leg crossing can actually reach). Either is a real extension
  of the crossing machinery, not a bug fix.

## One-line summary

`q q~ > t t~ q q~` is a genuine crossing of `q q > t t~ q q`, but the merged
`Q Qx` matrix element also carries the flavour-changing `d d~ > t t~ u u~`, whose
crossing needs a **final-leg reorder** the `I*(N+1)+J` encoding cannot express —
so the merged ME cannot be routed as a whole and keeps its own matrix element.

Qubitol:
I checked the latest messages, in particular those related to SIMD.
As far as I understood, in order to keep SIMD across the helicity amplitudes calculation, we need to evaluate the helicities at the beginning of each leg so that we are studying also the cross processes at the same time and we don't waste compute.

It seems reasonable to me as soon as we are unrolling the SIMD vectors only there, otherwise we are wasting the whole potential if we need to do that for each diagram (but I guess only external particles are involved right?).

Regarding the crossing encoding, I guess maybe we can allow for unordered leg check to understand if it is a crossing.
I wonder if this can happen also for crossing happening in the initial states.

oliviermattelaer:
### `acceptancetest_flavor` fixed: the recycled optim's helicity union is the SIGN map, not the GHREMAP permutation (db46ee28c)

`test_flavor_grouping_consistency_mlm` on `q q~ > q q~` was 105.8σ low (3706990 ± 10989 against the 3.7.2 reference 5184588 ± 2971, ratio 0.715) in the one configuration `apply_flavor_grouping=False` + `group_subprocesses=True`. It needs crossing **and** helicity recycling **and** a pruned config set: crossing on with `--hel_recycling=False`, or `--use_crossing=False` with recycling on, both give 5.189e6.

**Cause.** `crossgroup_helunion.dat` was carrying the wrong one of two base→base helicity maps. There are two because there are two consumers, and which one applies is decided by what the consumer can *apply* at run time:

```
sigma[h][k] = base_row[h][PERM[k]] * SGN[k]     permute AND sign-flip
tau[h][k]   = base_row[h][k]       * SGN[k]     sign-flip IN PLACE
```

`matrix<b>_orig.f` realises `sigma`: `APPLY_CROSSING_TABLE` permutes the whole NHEL table along with the momenta, and `CROSS_GHIDX` gates the shared `GOODHEL` through it. `matrix<b>_optim.f` cannot. The recycler bakes the helicity values into the HELAS calls as literals and the routine takes only `(PUSE, IC)` — IC carries the crossing's NSF sign flips, and nothing carries its slot permutation. Since `sigma = tau ∘ pi_unsigned`, optim row `h` is non-zero for a crossing iff `tau[h]` is good for the base, so the set to bake is `G_base ∪ tau(G_base)`. The file now holds `tau`.

Concretely, for the `u u~ > d d~` (base) → `u d~ > u d~` (router) crossing, which swaps slots 2 and 3: `PERM = 1 3 2 4`, `SGN = +1 -1 -1 +1`. HELAS builds each spinor from `nhel*nsf`, so flipping the NSF of slots 2 and 3 moves the non-vanishing condition of both currents from `h1 = -h2, h3 = -h4` (the base's s-channel chirality, rows `{1,4,13,16}`) to `h1 = h2, h3 = h4` (rows `{6,7,10,11}`). The sigma union `{1,4,7,10,13,16}` contains 2 of those 4, so the routed subprocess — all twelve `q q~' > q q~'` flavors, the t-channel half of the group — summed half its helicities.

Nothing about the *shape* of the wrong set gives it away, which is why the union looked verified: sigma and tau are both permutations, both involutions here, the sigma union is invariant under sigma, it covers every image of sigma, and the generated `matrix2_optim.f` sums over all its NCOMB rows. The set is closed under the wrong group.

**Also in the commit**

* `tau` is *always* a clean permutation (each leg's helicity states are closed under negation), whereas `sigma` need not be when the crossing swaps legs of different spin. The old "not a permutation" fallback wrote the identity as a keep-every-config marker, which stopped meaning that when the union replaced keep-all — it now adds nothing at all. That fallback is an explicit all-zero row instead.
* An identity `tau` (a crossing that moves no leg between initial and final, which genuinely needs no extra config) is no longer skipped: a non-empty `perms` list is also the marker that tells `gen_ximprove` the matrix element is shared by a crossing, which it needs in order to keep the C-parity de-duplication off.
* The colour-flow fix of 73b1aea7b is preserved, and not by luck. For `g g > u u~` ↔ `u u~ > g g` every leg crosses, so `tau` is the global helicity flip and both conditions (gluons opposite, quarks opposite) are invariant under it: `tau(G) = G`, the union collapses to the base's own good set, and `P1_gg_qq`'s optim is byte-identical to before. Where it does *not* collapse, a crossing base's optim necessarily holds rows that are dead for whichever member is calling, so AMP2 and JAMP2 (not the |M|² sum, which those rows cost nothing) are now gated on `TS(K) .NE. 0D0` — the same criterion the good-hel filter itself is trained on. Two new template holes carry it; they are empty for every non-crossing matrix element, so ordinary recycled output is unchanged.
* Noted on the way: the `INIT_MODE` good-hel print in `matrix<i>_orig.f` reports the **raw** loop index, which for a crossed flavor is a row of sigma-space. So a crossing base's reported good set is already `G ∪ sigma(G)` (6 rows here, not 4). That only ever makes the union a superset of what is needed, so it is not a correctness problem, but it is why the AMP2/JAMP2 gate is keyed on "is a crossing base" rather than on "the union grew".

**Validation**

| | |
|---|---|
| `test_flavor_grouping_consistency_mlm`, 4 settings vs 5184588 ± 2971 | True/False 5179400 ± 22594 · True/True 5194890 ± 15056 · False/False 5175765 ± 12239 · **false/True 5188690 ± 15750** (was 3706990 ± 10989) |
| that setting standalone, single core | 5.189e6 ± 1.6e4; `matrix2_optim.f` 6 rows `{1,4,7,10,13,16}` → 8 `{1,4,6,7,10,11,13,16}`, reproducing the 16-row keep-all run to 4 digits (5.193e6 both, same seed) |
| `TestMadeventCrossingBaseColorFlow` | OK — crossing-on and `--use_crossing=False` both 3.023e6 ± 1406 pb |
| `TestMadeventRouterColorSelection` | OK |
| `TestMadeventInclusiveCrossingXsec` | OK — 416.9 ± 2.4 routed vs 413.7 ± 2.6 independent (docstring reference 416.6 / 413.7) |
| `tests/acceptance_tests/test_standalone_cross_symmetry.py` | 65/65 OK |
| `TestGoodHelCParityDedup` | 4/4 OK |

`TestCrossingRecycledHelicityUnion` pins the map itself, run-free in 0.9 s: every recorded row must be the original row with the crossed legs' helicity negated in place, and the process must be one where sigma and tau actually differ. Pointing `_crossgroup_base_helsignmap` back at the permuted transform fails it immediately (`(1,-1,1,1) != (1,1,-1,1)`) instead of costing a four-build integration to notice.

PR #53, opened 2026-07-27
wire madspace logger through python logger, fix missing log messages

theoheimel:


oliviermattelaer:
Sounds good to me. We can merge

PR #58, opened 2026-08-05
ALOHA: merge multi-coupling routines, and make FD gauge routines self-contained

oliviermattelaer:
## Merged multi-coupling routines

`ALOHAWriterForFortran.write_combined` used to write `FFV2_4_3` & co as a **wrapper**: one call per Lorentz structure into a `<type>tmp`, then a component-by-component accumulation. Such a routine has no expression of its own — no contracted temporaries, no propagator `DENOM` — which costs a call plus an accumulation loop per use, and leaves nothing for the helicity-recycling batching to hoist out of its loop.

The combined routine is now computed like any other one: its body is the sum of the structures, each multiplied by its own coupling, sharing the momenta, the propagator denominator and the contracted temporaries. This re-uses `CombineRoutineBuilder`, i.e. the mechanism the loop (explicit-combine) routines already go through, and the writing goes through the standard `write()` path. The wrapper stays as `write_combined_wrapper` for the cases where the merge is not possible, and is reachable with `MG_ALOHA_COMBINE_WRAPPER` for A/B checks.

madmatrix already merged these (`explicit_combine=True` in `export_cpp`), so this brings the two backends in line.

## FD gauge

Two things made FD routines the exception, on both backends:

* `multiply_propagator_factor` was called at the end of every offshell V/S routine, recomputing for each helicity the part that only depends on the momentum (the 5-momentum `q`, the gauge direction `n`, `n.q`);
* structures acting on different spins — in FD gauge a leg is a massive vector in one structure and its Goldstone in the next — cannot share an ALOHA expression, so **every** FD combination was a wrapper (7 out of 7 for `w+ w- > w+ w-`).

Both are now written like a single-coupling routine: the propagator factor is emitted in the routine (`q`/`n`/`n.q` with the momenta, `js1`/`js2` and the update after the components), and the mixed-spin combinations are assembled term by term, sharing the momenta, the declarations, the contracted temporaries and one propagator factor. `define_gauge_dir` stays a call: it is a pure function of the momentum and sits in the invariant part.

For madmatrix the scratch wavefunction disappears with the wrapper, so the `FIXME: hack to avoid a bug in the FD code` that fed `aloha_obj_tmp`/`amp_tmp_fp` at the call site goes away too. `w+ w- > w+ w-` in FD keeps 17 calls in its DHELAS (all `DEFINE_GAUGE_DIR`), down from 48; its `HelAmps_sm.h` goes from 75 scratch references and 34 inner calls to 0.

## Three pre-existing madmatrix bugs found on the way

Independent of the above, in `aloha/template_files/madmatrix/helas_fd.h`:

1. **`vxxxxx`, scalar backends** — the FD Goldstone component of a longitudinal massive vector had the wrong sign (`+nsv*cI` instead of `-nsv*cI`, against fortran `vc%W(5) = -nsv*ci` and the python and SIMD implementations). Every matrix element with a longitudinal massive vector was ~0.7% off with `BACKEND=cppnone`, which is also what `cppauto` falls back to without SIMD. FD `u u~ > w+ w-` went from `5.0762980332466311e-03` to `5.1093407366301505e-03`, against `5.1093405664217962e-03` for the fortran standalone at the same phase-space point.
2. **`vxxxxx`, SIMD backends** — the FD block built its vector constants as `const fptype_sv one{1.}`, which sets the first lane and zero-fills the others. Everything selected with it was right in lane 0 only; the other lanes got `n = 0`, `nk = 0` and `-vmass/nk` = nan. `check_sa matrix` reads event 0, so single points looked fine, but a real multi-event run did not: `MeanMatrixElemValue = ( nan +- 0 )` becomes `( 0.339313 +- 0.306598 )`, and the multi-event mean now agrees between `cppnone` and `cppsse4`. The gauge direction there is now selected lane by lane with `fpternary` (dividing by the existing `ppDENOM`) instead of being combined arithmetically, since `nan*0` is `nan`; and the `|nhel|=1` choice is a plain `if`, `nhel` being a scalar.
3. **`define_gauge_dir`** — the scalar and SIMD branches disagreed on `n[3]` when the propagator has no spatial momentum. `n` is a gauge direction, so no matrix element changes (verified), but the choice should not depend on the backend; the SIMD branch now follows fortran, python and the scalar branch.

## Speed

Fortran standalone, best of 5 runs of `./check 1000 N`:

| process | before | after | |
|---|---|---|---|
| `p p > e- ve~ mu+ vm QCD=0` | 5.96 s | 5.73 s | +4.0% |
| `u u~ > w+ w-` | 13.04 s | 12.30 s | +6.0% |
| `e+ e- > e+ e- z` | 16.66 s | 13.81 s | +20.6% |
| `g g > t t~ g` (control, no combined routine) | 2.99 s | 2.99 s | 0.0% |
| FD `u u~ > w+ w-` | 2.15 s | 1.54 s | +39.6% |
| FD `w+ w- > w+ w-` | 170.8 s | 66.5 s | 2.57x |

madmatrix FD `w+ w- > w+ w-` (`check_sa.exe perf`): 24053 to 49487 evt/s, 2.06x. ALOHA generation costs 3-8% more on the processes measured.

## Validation

* **Per routine, every component** (20 pseudo-random inputs, matching and rejected flavour configurations): `FFV2_4`, `FFV2_5`, `FFV1_2_4`, `FFS1_2`, `VVVV1_3_4` (3 couplings), plus the `C1`, `P0` and `P1N` tags — max 1e-15 relative, i.e. the reassociation of the sum. Same check on the 48 FD routines of `w+ w- > w+ w-` and `e+ e- > e+ e- z`.
* **Matrix elements**: 16 process/model configurations A/B'd through `check_sa`, worst 7e-15 — including `p p > z j j` (7 subprocesses, 30 flavour matrix elements), MSSM, heft and decay chains. FD: bit-identical for `u u~ > w+ w-`, `w+ w- > w+ w-`, `u u~ > z z`, `e+ e- > e+ e- z`, 1 ulp for `e+ e- > w+ w- z`.
* **madmatrix**: bit-identical on the FD processes and byte-identical output for non-FD; both `cppnone` and `cppsse4` built and compared, in `matrix` and `perf` mode; `vxxxxx` driven directly on a SIMD register holding a boson at rest next to a moving one, every lane finite and equal to the scalar backend.
* **NLO**: `p p > e+ e- [QCD]` generates and compiles, including the quadruple-precision `MP_FFV2_4_3`; loop routines keep their own explicit-combine path.
* **Tests**: `test_madevent_ufo_aloha`, `test_madevent_ufo_aloha_merged`, `test_mg7_ufo_aloha`, `test_standalone_madevent_consistency` (9), `test_madevent_flavor_zud`/`zjj`, `test_madevent_merged_flavor_uq`, `test_single_qcd_order_mg7`, `test_group_subprocess_mg7` all green. The `FFV2_4_3` reference text in `check_aloha_file` is updated to the merged routine.

Not covered: `cppavx2`/`cpp512*` and CUDA (no GPU here), and the plain `ALOHAWriterForCPP` (`standalone_cpp`) backend, which still emits the FD wrapper.


oliviermattelaer:
To my point of view this is good to be merged (but this is a quite limited and independent merge).

Qubitol (approved):
Yes, let's merge.
I'm just confused about some files with the weird names `%FD_madmatrix%src%HelAmps.....`, but I assume/understood those are used for tests.

oliviermattelaer:
Yes those are reference file to detect unexpected drifting of the code between version (which also document the change in the generated code)

PR #60, opened 2026-08-08
Cheaper colour sum: DDM colour basis, Kleiss-Kuijf flows, and the same for madmatrix

oliviermattelaer:
Makes the colour sum of a matrix element cheaper, and gives the C++/cudacpp
backend the colour-flow machinery the Fortran one already had. None of it
changes a physics result.

## What is in here

**A (n-2)! colour basis for multi-gluon processes.** The Jacobi identity lets
any fully adjoint colour factor be written on the (n-2)! Del Duca-Dixon-Maltoni
half-ladders instead of the (n-1)! traces. The reduction walks each diagram's
colour string as a tree and expands every subtree hanging off the leg-1-to-leg-n
spine as a nested commutator, at 2^(n-2) per diagram. New
`set color_basis auto|trace|ddm`; `auto` takes it for the exporters that declare
`support_ddm_color_basis` -- Fortran standalone, grouped and non-grouped
madevent, and madmatrix. Processes that are not fully adjoint fall back on their
own, per amplitude.

**Kleiss-Kuijf colour flows.** A colour flow still has to be picked among the
(n-1)! trace structures, but those JAMPs need not be built from the amplitudes:
`J_trace(rho) = sum_sigma c_{sigma rho} J_DDM(sigma)`, using the expansion
coefficients the colour matrix already computes. At `g g > 5g` that is **3840
terms instead of 231840**, 3.0 us against 22.4 us per helicity. LHE colour
flows are unchanged.

**The same treatment for the madmatrix (C++/cudacpp) backend**, which had none
of it: the DDM basis and the KK flows (`ncolor` and `ncolor_flow` are now two
different numbers, and every jamp2 buffer moves to the second), a host BLAS
colour sum, and the JAMP common-subexpression and orbit-equivariant searches,
which are now shared with the Fortran exporters through the new
`madgraph/iolibs/jamp_optimiser.py` rather than duplicated. The Fortran output
is byte-for-byte unchanged by that sharing.

**A batched BLAS-3 colour sum for the C++ host path.** The colour matrix does
not depend on helicity, so every good helicity is a column of one right-hand
side and the sum is two SYMM calls instead of a triangular loop per helicity.
Accelerate's SYMM does about twice the flops and still wins 9x on its own: the
colour sum drops from 31% of the matrix element to 7%, **1.30x** end to end on
`g g > g g g g` and **1.13x** on `g g > t t~ g g g`. Below the `ncolor`
threshold, or with no BLAS to link, the generated files are character for
character what they were.

## Numbers

`g g > 6g`, Fortran standalone, per phase-space point, seconds:

| | point 1 | 1-10 |
|---|---|---|
| trace basis | 2.22 | 2.19 |
| DDM basis | **1.24** | **1.20** |
| DDM at -O2 | **1.01** | **0.99** |

1.82x on the first ten points and 1.87x in generation, and it is what makes the
first point cheap. Steady state is a different story and was measured when the
reflection folding was still in: 0.544 s trace against 0.554 s DDM, i.e. a
wash, with the folding on the trace side. That measurement has not been redone
since the folding came out, so the trace figure there is now optimistic.
Either way `GET_AMP` is 85% of the steady-state time, which nothing here
touches.

`CPPProcess.cc` for `g g > g g g g`: 512 -> 269 kB, with `ncolor` 120 -> 24.

## Correctness

- `f.complex_conjugate` was silently wrong. The inherited implementation
  reverses the indices, which is right for `Tr` but flips the sign of the
  totally antisymmetric `f`. It was latent because `full_simplify` removes every
  `f` before the trace-basis colour matrix is built; with the DDM basis the
  elements are products of (n-2) f's, so |M|^2 came out **negative for an odd
  number of gluons**. Fixed.
- The mg7 exporter built the `icolamp` mask on the colour-sum basis while the
  C++ selection reads it over `ncolor_flow`, i.e. a `[3][2]` array read 6 wide.
  |M|^2 never saw it; colour flows did, by 134 sigma on one flow of
  `g g > g g` over 3 x 100k events. Fixed, and the generated `icolamp` is now
  byte-identical to the trace-basis one.
- |M|^2 matches the trace basis to 1e-16 relative at every multiplicity and at
  four phase-space points for `g g > 5g`; `g g > 6g` is bit-identical.
  `g g > g g g g` in madmatrix: 1.5929925846563478e-04 against
  1.5929925846563475e-04.
- `g g > g g` through madevent with the same seed gives the identical cross
  section and **identical colour flows on all 8000 particles of 2000 events**.
- The KK-rebuilt trace JAMPs match the directly built ones to 2e-16, all 120 of
  them at `g g > g g g g`.

## What was tried and dropped

Two things advertised in earlier revisions of this PR are no longer in it, both
removed on measurement rather than on taste:

- **Colour-reflection folding** (`9f8b1b334`). Summing |M|^2 over one flow per
  reversal pair only ever applied to a pure gluon process on the trace basis:
  the DDM half-ladders are not closed under reversal, so every exporter that
  folded also declares `support_ddm_color_basis` and none of them sees a
  foldable basis in `auto` mode. What was left was an explicit
  `set color_basis trace`. The DDM basis is faster than the folded trace one
  was anyway (3.59e-4 s against 3.78e-4 s per 32-event madmatrix batch at
  `g g > 3g`).
- **The batched BLAS-3 colour sum for Fortran**, madevent (`b2b6d13ef`) and
  standalone (`bfc5a0814`). `g g > 5g` 30.10 s -> 30.30 s and `t t~ 3g`
  9.68 s -> 9.79 s: run-to-run noise. Once the DDM basis has taken the colour
  sum down to a few percent of the matrix element there is nothing left to win.
  The C++ host BLAS above is kept, and with it the `blas_*` probe, which is now
  its only caller.

## Notes for review

- The GPU path is generated but **not compiled**: there is no CUDA or HIP
  toolchain on the machine this was developed on. That covers the device side
  of the colour sum.
- The IOTest goldens under `tests/input_files/IOTestsComparison` are
  regenerated on this branch (14 files) and the IOTests pass.
- Merged with `main` at `cc0c510aa`. Six conflicts, all resolved; the only
  non-mechanical one is `set_channels_colors_map` in `export_mg7.py`, where
  main's leading-Nc `icolamp` filter and this branch's move to the flow basis
  had to be combined -- the resulting `icolamp` is byte-identical to main's.
  Post-merge the unit suite matches `main` failure for failure, with 32 extra
  tests from this branch, all passing.


oliviermattelaer:
Full unit suite on this branch: **917 tests, 3 red.**

| test | status |
|---|---|
| `testIO_export_matrix_element_v4_madevent_group` | **new here** — golden drift, the madevent template now emits `NCOLOR_FLOW` |
| `testIO_UnitProcOutputIOTests` | pre-existing — fails identically on the parent branch (checked by reverting this branch's files and re-running) |
| `test_DensityMatrixObservables22` | pre-existing — needs scipy, which is not installed here |

So one genuine reference to regenerate, and it is a reference rather than a
behaviour change: `ncolor_flow` equals `ncolor` whenever no smaller colour-sum
basis is in play, so the generated file is numerically what it was.

`testIO_sqso_uux_uuxuuxx` also fails on the parent branch but passed in this
run — it is a border effect between test files, not something this changes.

oliviermattelaer:
Two updates.

**The batched BLAS-3 colour sum is gone from madevent** (`b2b6d13ef`). Re-measured
after the DDM colour basis landed, it bought nothing: `g g > 5g` 30.10 → 30.30 s,
`t t~ 3g` 9.68 → 9.79 s, both marginally the wrong side of unity. It does engage
(both sit at `NCOLORFOLD` 120); there is simply nothing left to win once the colour
sum is a few percent of the matrix element.

Scope is exact rather than approximate: as line multisets the commit removes the
291 lines `7fd3a33e3` added and restores the 25 it removed, and no others.

Deliberately kept, because they measure differently:

- **the madmatrix host BLAS** — 1.02x at `g g > 5g`, 1.15x at `t t~ 3g`, **1.33x at
  `t t~ 4g`**, growing with the process. It shares `blas_is_available` and
  `blas_min_ncolor` with the fortran side, so those helpers stay.
- **the fortran standalone batch** (`990cd7f69`) — untouched, and not covered by the
  sweep above. The only evidence is an earlier seven-leg 1.37x on the trace basis and
  1.07x on the DDM one. Left in place pending its own measurement.
- **the device cuBLAS/hipBLAS path** — upstream cudacpp, untouched.

Generated madevent output is byte-identical to a batch-off build of the previous tip
across grouped/non-grouped x hel_recycling on/off; `standalone` and `standalone_mg7`
are byte-identical to the previous tip with BLAS on. A `p p > t t~` run at equal seed
gives the same cross section to every stored digit and bit-identical colour flows.

**Correction to the test baseline** in my earlier comment: it is **923 tests, 3 red**,
not 917. The count drifted before this work — it is 923 on both branches — so the
earlier figure was stale. The three reds are unchanged:
`test_DensityMatrixObservables22` (wants scipy),
`testIO_export_matrix_element_v4_madevent_group` and `testIO_UnitProcOutputIOTests`.

Worth noting for reviewers: no test or IOTest golden references BLAS anywhere, so this
removal had no coverage to lose, and the `blas_color_sum` run-card parameter was hidden
and never reached a shipped `run_card.dat`, so removing it cannot invalidate an
existing card.

oliviermattelaer:
**The fortran standalone batched colour sum is gone too** (`bfc5a0814`). That leaves
the madmatrix host BLAS as the only batched colour sum in the tree, which is the one
that measures a real gain (1.02x / 1.15x / **1.33x** as the process grows).

Scope was checked by attributing every removed line to its origin commit with
`git blame`, which is stronger than a diff review: 226 lines, from
`990cd7f69` (180, the batch itself), `20446e4b2` (20, madmatrix helpers *relocated*
not deleted), `7487ad5ba` (20) and `b33129eda` (6). **No line from a non-BLAS commit.**

The 20 lines from `7487ad5ba` are the `GET_JAMPF` call that lived inside the batch
branch, and removing them is correct: that call existed only because the batch never
enters `MATRIX`. Verified after the fact — standalone `g g > 5g` still defines and
calls `GET_JAMPF` exactly once through the normal path, so the Kleiss-Kuijf colour
flow rebuild is intact.

**Where the BLAS probe lives now.** `blas_is_available`, `blas_available_flags` and
`blas_min_ncolor` were on `ProcessExporterFortran`, but with both fortran batches gone
their only caller is the C++ backend. They now sit on `OneProcessExporterMadMatrix`
next to `cpp_blas_wanted_for`, and the cross-package import is gone —
`ProcessExporterFortran` and `misc` carry zero BLAS symbols. Confirmed working:
`cpp_blas_wanted_for` is False at ncolor 6 and True at 120, and
`blas_available_flags()` still reaches the generated `madmatrix.mk`.

Note for anyone tempted to move such helpers to `madgraph/various/misc.py`: that file
is copied into **every madevent process directory** as `bin/internal/misc.py`, so
anything added there ships into every run dir and moves madevent's generated output.

Generated output: `standalone` differs from a `blas=False` build of the previous tip
by exactly the `BLASLIBS` line and three blank lines (the placeholders sat on lines of
their own, so `blas=False` could never fully suppress them — it can now).
`madevent` and `standalone_mg7` are byte-identical to the previous tip, the latter
both below and above the ncolor-100 threshold.

|M|^2 moves by ~1 ulp where the batch used to engage — the scalar loop sums the terms
in a different order. `g g > 5g`: 6.6739867626784560e-07 against 6.6739867626784539e-07.
Below the threshold the two builds are bit-identical.

Tests unchanged: **923, 3 red** (`test_DensityMatrixObservables22` wants scipy, plus
the two IOTest goldens). Verified on both the previous tip and this one.

Unrelated but worth flagging for reviewers: `output matchbox` currently fails with
`TypeError: get_JAMP_lines() got an unexpected keyword argument 'proc_prefix'`. It
fails identically on the previous tip, so it predates this branch's colour work.

oliviermattelaer:
So I have remove the BLAS-3 part and the colour reflection which were both useless as soon as i implemented the DDM method. This one is good to be merged !

Qubitol (approved):

PR #63, opened 2026-08-12
MadSpin: generate decay events with mg7, and cut the overheads around them (2.0x at 10k, 4.4x at 100k)

oliviermattelaer:
Speeds up MadSpin by moving decay-event generation off Fortran MadEvent and onto
the mg7 (madmatrix/madspace) engine, then removing the overheads that dominated
what was left.

`p p > t t~` with both tops fully decayed (`t > b w+, w+ > all all` and the
charge conjugate), 18 cores:

| events | before | after |
|---|---:|---:|
| 10 000 | 80.5 s | ~40 s |
| 100 000 | 283.5 s | ~64 s |

Every number below is a back-to-back measurement against the immediately
preceding commit. This machine's load drifts by up to 40% between runs, so
same-session pairs are the only trustworthy comparison, and unrelated phases
were checked for movement each time.

## Why decay-event generation moved

It was 80.5% of a run: 64.0 s of 79.5 s at 10k events, against 4.9 s for the
accept/reject loop those events feed. Most of that was fixed cost. A MadEvent
pool refill costs ~12 s of survey/refine/combine whatever its size — the
baseline paid 23.7 s to produce 4708 events.

mg7 could not do this at all: madspace rejected anything but two incoming
particles, and the launcher is built around beams, PDFs and a luminosity.

- **madspace** accepts a single incoming particle. A decay is the s-channel
  cascade the 2 → n mappings already build, with the t-channel chain absent and
  the root virtuality fixed rather than sampled, so this is mostly about not
  assuming the incoming count is 2. Validated against analytic phase-space
  volumes: 1 → 2 massless and massive exact to 1e-10, 1 → 3 massless to 0.02%.
- **The mg7 exporter** was writing garbage for decays without complaining —
  `incoming = [pdg, None]`, and an outgoing offset of 3 that dropped the first
  final-state particle. `t > b w+, w+ > all all` came out as
  `outgoing = [81, -81]`, with the b silently gone.
- **The mg7 launcher** gained a decay mode: no PDFs, scale fixed at the decaying
  mass, `dGamma = |M|^2/(2M)` in GeV rather than a cross section in pb, and a
  constant alpha_s taken from the param card rather than demanding an LHAPDF set
  to evaluate a coupling.
- **Kinematically closed channels are dropped.** `w+ > all all` enumerates
  `w+ > t b~`, `b W+ Z` and `b W+ h`; their width is exactly zero but the mapping
  has no physical point to return, so it produced NaN momenta and poisoned the
  whole integral.

Result: decay-event generation 1.68x, refills **63x** (mg7 returns exactly the
number of events asked for, so pools rarely run dry), total **1.98x**.

## The rest

- **Matrix elements build in parallel.** `misc.compile` defaults to `nb_core=1`,
  and a subprocess has only ~5 objects so `make -j` alone cannot fill the
  machine. Subprocesses now build concurrently with the job budget split between
  them: **3.30x** on the launcher phase.
- **Decay pools are numpy, not LHE text.** mg7 writes the same fields as a
  structured array; MadSpin builds its events from that instead of parsing.
- **gzip.** `misc.gzip` used the module default `compresslevel=9` for files under
  256 MB while shelling out to `gzip` (level 6) above it — the same data
  compressed differently depending on its size, with the common case on the slow
  branch. Level 6 costs 4.5 s against 18.6 s on a 172 MB LHE and buys 4%. It also
  read the whole file into memory as a str and encoded it; it streams now.
  Separately, MadSpin gunzipped its read-only input and repacked identical
  content at the end. **4.20x** on that phase, and peak RSS halved (781 → 393 MiB).
- **Decay loop.** Per-trial re-reads of the param card and the process directory
  removed; `Event.boost` no longer allocates two FourMomentum objects per
  particle (**4.98x**, within 8% of the floor set by attribute access); `trace`
  and `scalar_multiplication` no longer call numpy to add two numbers.
- **A batched density entry point** (`PY_GET_DENSITY_BATCH`), used by the
  max-weight scan.

## Physics

Unchanged, and checked directly rather than inferred from the total. Partial
widths against Fortran MadEvent at 100k:

```
t  > b w+, w+ > all all   1.4600457(7474)  vs  1.459496(13454)   0.36 sigma
t~ > b~ w-, w- > all all  1.4615953(7442)  vs  1.460260(2380)    1.71 sigma
```

with the mg7 errors 2.4x smaller for half the wall time. The max-weight scan
batching is bit-identical trial by trial (1475/1475, max|diff| exactly 0), as is
the first boost rewrite (39976 momentum components). The later changes are
summation-order only: complex64 epsilon for the density algebra, 1.2e-12 for the
boost.

## Tests

`test_madspin -p U` (23), `test_lhe_parser -p U` (24), the madspace suite (1482,
up from 1469), and four MadSpin acceptance tests. Two new regression tests:
`test_output_mg7_decay_subprocess_metadata` (fails on the pre-fix exporter) and
`test_decay_topology.py` for the 1 → n mappings.

Three MadSpin acceptance tests fail on the development machine both before and
after this branch — the production step crashes in systematics on lhapdf's
broken Python 3.14 bindings, before MadSpin is reached. Verified against the
pre-change files.

## Also worth knowing

- `decay_generator` in the madspin card selects the backend; it defaults to
  `mg7`, with `madevent` as a one-line fallback and forced for gridpack mode.
- The mg7 run card has no seed, so decay pools are not reproducible run to run.
  MadSpin's own accept/reject RNG is still seeded. Worth adding.
- `MG_LHE_TIMERS=1` enables the LHE parser timers, which existed but could not be
  switched on.
- `tests/parallel_tests/madspin_benchmark.py` reproduces every number here.
- Two pre-existing bugs in the shared test factory were fixed on the way: the
  `set` lines were written after the `done` that starts the run, so nevents,
  iseed and the beam energies never applied; and a crashed production run was not
  detected.

## What is left, with measurements

The decay loop is now ~42 s of ~64 s at 100k, and it is not the matrix element:
per trial, momenta extraction is 23 us, weight assembly 22 us, DensityMatrix
construction 7 us, and the Fortran matrix element 9 us. The largest remaining
item is vectorising `get_momenta`/`get_pdg` (~12% of the run). Porting the
density to the madmatrix C++ backend is bounded by Amdahl at 4.7% whatever the
vector width, and madmatrix has no density support today.


oliviermattelaer:
Hi Theo,

While this is mainly moving madspin backend from madevent to madspace.
I had to activate the madspace support for 1>N matrix-element that you might want to check before approving.

The rest is some additional validation for the mg7 paper to say that madspin2 is faster in mg7 compare to the madspin2 paper (soon to be released).

Olivier

theoheimel (approved):
The madspace changes look good to me

oliviermattelaer:
## `test_short_madspin_ttbar` / `test_short_madspin_singletop`: blocked on #51

Both fail on the same assertion, comparing two spinmodes inside the `run_onshell` family:

```
ttbar:     onshell_decay_chain = 71.4322 pb  vs  onshell_density = 71.1386 pb   rel = 0.411%
singletop: onshell_decay_chain = 25.6976 pb  vs  onshell_density = 25.5744 pb   rel = 0.479%
tolerance: 0.100%
```

**This is not a physics discrepancy.** It is a loss of reproducibility, and the tests are right to notice.

### Cause

The madevent path seeds each decay-event pool (`interface_madspin.py`, `run_card["iseed"] = self.seed; self.seed += 1`). Since the factory writes `set seed 42` into every mode's card, every spinmode got a **bit-identical** decay pool, so an identical partial width and an identical BR. The only residual difference between modes was the unweighting — which is what the 0.1% tolerance was measuring.

mg7 has no seed. Running MadSpin twice per backend with identical cards on the same production sample:

| | partial widths | σ×BR |
|---|---|---|
| madevent, run 1 | 0.3237400, 0.9708000 | 71.25510 |
| madevent, run 2 | 0.3237400, 0.9708000 | 71.25510 |
| mg7, run 1 | 0.3248777, 0.9719431 | 71.32298 |
| mg7, run 2 | 0.3237563, 0.9676568 | 71.24660 |

madevent reproduces bit-for-bit; mg7 does not.

### The failures are within one sigma

madspace quotes the width errors itself — 0.282% and 0.283% on the two decays. BR is their product, so per run the relative error on BR is 0.40%, and comparing **two independent runs** gives an expected 1σ spread of **0.567%**.

Observed: 0.411% and 0.479%, i.e. both under 1σ. The test asks for 0.100%, which is **0.18σ**, so its pass probability is roughly **14%**. `test_short_madspin_zz` passing is luck, not a distinction.

### Not specific to MadSpin

madspace seeds nothing — `std::random_device` at seven sites across the CPU runtime, GPU runtime, event generator and MLP. Same directory, same card, `p p > t t~ j`:

```
run 1: integral: 1006(11)
run 2: integral:  987(11)
```

This branch did not create that gap. It made a test notice it, by swapping a seeded generator for an unseeded one inside a comparison that assumed determinism.

### Why we need #51

**#51 ("Implement reproducible event generation") is the fix, and this PR should wait for it.** Its stated goals — full reproducibility of generation in regular and gridpack mode, independent of worker thread count — are exactly what is required here. Once decay pools are reproducible, the modes again share an identical BR by construction and the 0.1% tolerance goes back to measuring the unweighting.

Worth stressing, because it is the tempting shortcut: **please do not fix this by loosening `rel_tol`.** The assertion's own docstring calls it "the strict invariant catching real bugs". Relaxing it to ~2% (3σ) would make the tests green while leaving them unable to catch a genuine 0.5% error in the unweighting normalisation — precisely the class of bug it exists for. Loosening removes the test's power rather than restoring it.

### Until #51 lands

The two tests will keep failing about six runs in seven. The interim option, if this PR needs to land first, is to flip `decay_generator` back to `madevent` by default and document mg7 as opt-in (`set decay_generator mg7`) — that restores determinism and keeps the tests meaningful, at the cost of the speedup being off by default. Preference is to wait for #51 rather than ship that.

Also worth adding once #51 is in: the comparison would be more honest as an n-sigma test against the combined MC error rather than a fixed relative tolerance, since a fixed tolerance is only valid while the two sides are correlated by construction. That is a follow-up, not a substitute for #51.

PR #66, opened 2026-08-14
keep a zero-density flavour channel out of the MadNIS densities

oliviermattelaer:
A flavour channel whose parton densities are all exactly zero produces
`det = 0` in `kernel_sample_discrete_probs`. That is the right answer for the
event weight, which it zeroes, and the wrong answer for everything that treats
a det as a jacobian. The MadNIS training path does exactly that with **both**
of its densities:

- `q`, the sampling density, is `norm/det` (`Integrand::build_common_part`), so
  a degenerate draw gives `1/0 = +inf`.
- `g`, the model density, is the inverse det (`IntegrandProbability`), i.e. the
  zero probability itself. `madnis_variance` computes `f/g - mean`, so the row
  is NaN whatever `q` is — **fixing only the `+inf` would not have been
  enough.** `madnis_softclip` also returns NaN from `inf * asinh(0)`.

Today this is contained only by the explicit `w != 0.` test in `op_unweight`,
which keeps zero-weight points out of events. Nothing structural stops the
column from reaching MadNIS, which consumes `adaptive_prob` directly.

## The fix

A degenerate discrete distribution carries no information: the index that comes
back was not a choice, so the density of the row is the marginal over the
discrete step, which is **1**. That is also the neutral factor a row failing the
cuts already contributes to the column — the right analogy, since a phase-space
point where every parton density of the channel vanishes is a zero of the
integrand exactly like a cut failure, and MadNIS should learn to avoid it rather
than see an infinity or a NaN.

- **Forward** (`integrand.cpp`): the det is split with `max(det, 1)` for the
  density and `min(det, 1)` for a 0/1 mask multiplied into the weight. `det = 1/p`
  is `>= 1` for any real draw and `0` only for the sentinel, so the split is exact
  — both are bit-for-bit identities on a regular det — and needs no new opcode.
- **Inverse** (`discrete.hpp`): the degenerate det returns the same marginal `1`
  instead of the zero probability, and the backward guard moves from the norm to
  the selected probability so the gradient is zero wherever the forward det is
  now a constant.

The weight mask matters on its own. Under `madnis_training` the det is kept out
of `weight` (it is applied separately, to `full_weight`), so the loss would
otherwise see a non-zero `f` against a finite `q` for a row whose event weight is
zero — and the option a degenerate draw returns can be one the active-flavour
mask switched off, whose parton density is not zero at all. That is the same trap
that sank the uniform-fallback attempts inside the kernel. With the mask, `f/q`
reproduces the event weight the integrator uses, which is 0.

## Measurements

Real mg7 integrand, `p p > go go` with `NNPDF40MC_lo_as_01180` pinned,
9 channels x 1e6 points:

| | non-finite `adaptive_prob` | zero-weight rows |
|---|---|---|
| before | 359696 / 9e6 (4.0%) | 0 |
| after | 0 | 360331 (= the degenerate rows) |

Every one of those rows has a `pdf_prior` that is zero in every flavour. After
the change they carry `adaptive_prob = 1` and `weight = 0`.

Cross-sections (50000 events): **5.0144 +- 0.0039** with the default PDF
(reference 5.024, 0.19%) and **3.7890 +- 0.0030** with NNPDF40MC pinned, i.e.
the PDF that makes the degenerate rows happen (reference 3.786, 0.08%).
Nothing outside `madnis_training` changes: `max`/`min` are identities on a
regular det, and in the plain integration path `adaptive_prob` is not even an
output.

`madspace/tests/test_discrete_sampler.py` pins the invariant the split relies on
(det is 0 or `>= 1`), the inverse det never being 0, and the three `MadnisLoss`
kernels staying finite on a degenerate draw. madspace pytest: 1468 passed, 4
failed (the pre-existing scipy ones). `test_generation_from_file_1_mg7` and
`test_relaunch_switch_defaults_mg7` pass.

**Not exercised:** the training loop itself needs torch, which is not installed
on the machine this was developed on. The training-mode `Integrand` and the
`MadnisLoss` kernels were built and run directly, which covers the columns and
the loss arithmetic, but not `MadnisTraining::train_step` or an optimizer step.

## Note on the first commit

The first commit is a cherry-pick of 621c9c907 from `feat-big-renaming`
(kernel hunks only) — this PR needs the `det = 0` sentinel it introduces and
should not have to wait for that branch. Its test hunk is deliberately left
behind: it re-references `test_generation_from_file_1_mg7` to NNPDF40MC, which is
only the mg7 default on that branch. `feat-big-renaming` touches nothing else
under `madspace/`, so **rebasing** it after this lands drops the duplicate
cleanly and leaves just that test hunk where it belongs.


theoheimel:
Claude's fix for the discrete sampling bug was quite convoluted. I replaced it with something simpler. This still fixes the `p p > go go` for me, but feel free to double check. I also noticed we had no handling for negative PDF values in the flavor sampling, so I added an absolute value there. Otherwise this is ready to be merged from my side.

oliviermattelaer:
Checked it out, rebuilt, and re-ran the measurements against your version — the simplification does fix the issue, and it fixes it better than the sentinel in one respect. One residual defect, small but real.

**Fixed.** Never selecting a zero-probability option means `det` is always `>= 1`, so the `1/0` I was splitting is gone at the source:

- `adaptive_prob`: **0 non-finite out of 9e6** rows (`p p > go go`, `NNPDF40MC_lo_as_01180` pinned, 9 channels x 1e6), against 359696 with only the previous guard in place.
- `g`: the `prob_norm == 0 -> det = 1` branch in the inverse keeps `f/g` out of `0/0`, so `madnis_variance` is finite too. All three `MadnisLoss` kernels finite on a degenerate row.
- `p p > go go`, 50000 events: **3.7836 +- 0.0029** with NNPDF40MC (reference 3.786) and **5.0216 +- 0.0038** with the default PDF (reference 5.024). madspace pytest: 1470 passed, 4 failed (the pre-existing scipy ones).

Excluding a zero-probability option from the walk outright is also strictly better than what I had for case (B) — it can no longer be selected at all, instead of being selected and then patched up — and the `abs()` on the prior is a real improvement I had left alone.

## The residual defect

The all-zero fallback returns index 0, and the integrand gathers the parton density of the *selected* flavour from the **unmasked** values (`integrand.cpp`, `pdf = fb.gather(fb.gather_int(flavor_id, indices), pdf)`); the `active_flavors` mask is applied only to the sampling prior a few lines above. For `p p > go go` the per-channel masks are disjoint — chan 1/2 = d d~, chan 5/6 = u u~, chan 7/8 = c c~ — and a row is degenerate precisely because *that* channel's flavour has an exactly-zero density while d/u/s do not. So index 0 is a flavour the channel does not carry, and its density is not zero:

| | chan 5/6 (u u~) | chan 7/8 (c c~) |
|---|---|---|
| degenerate rows | 14813 / 1e6 | 165074 / 1e6 |
| flavour index returned | `[0]` = d d~ | `[0]` = d d~ |
| outside `active_flavors` | 100% | 100% |
| non-zero weight | 100% | 100% |
| `adaptive_prob` there | 1 | 1 |

All 360504 degenerate rows pick a flavour outside the channel, with sampling density 1 and a non-zero weight — a double count of d d~, which channels 1/2 already integrate. The channel's own contribution in that region is exactly zero, so all of what those rows contribute is spurious.

It is **numerically tiny here**: 0.017% of the weight of channels 7/8, 4e-7 of the total, max weight 0.5% of the channel max — far below the 0.08% MC error, which is why the cross-section above is untouched and why such a row would essentially never survive unweighting. But the suppression is not structural: it comes from those rows living at large x where the luminosity is small, and it scales with how much of a channel's phase space is degenerate.

## Suggested fix

One multiplication in `Integrand::build_channel_part`, enforcing on the weight the invariant the mask already expresses for the prior — *a channel only integrates the flavours it carries*:

```cpp
// mask = fb.gather_vector(chan_index, _active_flavors_mask), as used for the prior
weights_after_cuts.push_back(fb.gather(flavor_id, mask));
```

With your kernel a masked-out flavour can only be selected on an all-zero row, so this is exactly 1 on every other row and nothing else moves. It lands on `f = 0, q = 1, g = 1` for the degenerate rows — the same end state as my patch, without any sentinel, and it closes the wrong-flavour case however it is reached rather than only the ones we have thought of.

The new tests don't cover this; worth one asserting that the flavour returned is inside `active_flavors` or the weight is zero.

Happy either way on who writes it — say the word and I'll push it on top of your commits, or leave it to you.

To reproduce: build a training-mode integrand from a generated run dir with `NNPDF40MC_lo_as_01180` pinned in the run_card (`subproc.build_integrands(ps, madnis_training=True)`, no torch needed), run a batch, and compare `discrete_flavor_index` against the channel's `active_flavors` on the rows where `pdf_prior` is zero in every flavour.

theoheimel:
I don't think there is an issue here, since the channel weight returned by the matrix element should be zero when called for an inactive flavor and therefore the final weight should be zero even before the suggested fix.

oliviermattelaer:
ok let's merge this then

PR #72, opened 2026-08-18
MLM merging

theoheimel:
## Overview

This branch implements MLM merging in madspace. The core part is a clustering algorithm that takes momenta as its input and returns:
- renormalization scale
- factorizations scales
- scales assigned to outgoing jets
- a diagram index that will be used to assign color flows, resonances etc
The first two affect the cross section and PS distributions, whereas the last two are relevant for the LHE output.

The two central parts are:
- `madspace/src/phasespace/mlm_clustering.cpp`: takes the diagram topologies for the given subprocess as input and compiles them into an efficient binary representation of the possible clustering histories. This representation will also nicely generalize to Amplicol.
- `madspace/src/kernels/mlm.hpp`: the actual compute kernel run during event generation for CPU and GPU. Based on the previously compiled state machine, it determines a clustering and then assigns the scales. The functions `djb_clus`, `dj_clus`, `compute_scale`, `update_momenta` are directly ported from the NLO fortran code. The actual clustering logic happens in `mlm_clustering`. The actual scale computation (after a clustering has been determined) is mainly based on my conversations with Rikkert when he was at CERN, and should be similar to what the NLO code does.

## Things to discuss

Rikkert might have opinions here. I don't really know the physics well enough...

- **how to deal with resonances**: MG5 (both NLO and LO) does that based on the SDE integration channel which is not an option with madspace/madnis. The current implementation does it by prioritizing resonant clusterings over non-resonant ones using a cutoff. This might not be ideal and result in non-continuous PS distributions.
- **what should the actual scale look like**: The current version resembles the NLO one (without the Sudakov correction). Some things are quite different from LO. For instance, the factorization scale is assigned to be the softest clustering, no matter if it is actually linked to the initial state.

## Missing pieces

- [ ] this branch has not been touched in a while, so we should probably merge some of the recent changes
- [ ] `madspace/src/phasespace/mlm_clustering.cpp`: right now the metadata whether a given vertex is QCD, an output is a jet, propagator masses etc. does not get populated. Instead massless QCD is assumed. In principle all the necessary information should be there, so this should be trivial to change
- [ ] pass on the selected diagram index to the matrix element. There is a currently unused UMAMI input key for that, but it will require some changes on the madmatrix side to support a per-event diagram input (@Qubitol has worked on this). Not so urgent since this only affects the LHE output, not the actual scales. So maybe we can do this once we have merged some other stuff (especially @stloufra's backend separation)

## Validation

Here is a three-step plan. The challenging part of validation is that it is not always clear what the baseline is (see things to discuss), also given that to some extent there is no right and wrong here.

### Unit tests

- check clustering result at the level of individual phase-space points. A simple "smoke test" is already there
- baseline suggested by Rikkert: scales extracted from a MG5 NLO run, but without the Sudakov corrections. Will require some hacking. Goal: identical output to NLO (if this is what we want...). Problem: works for ren/fac scale, but we don't really have a baseline for the scales assigned to the jets since this works differently in FxFx and there is only one scale passed to the shower

### Parton level

- do cross sections look reasonable compared to MG5 MLM?
- are the scales reasonably distributed?
- what does reasonable even mean? is this something we can systematically test?

### Shower level

- compare merged observable histograms after shower
- do we get smooth distributions?
- are there processes where things are likely to fail?


Qubitol:
Thank you.

The changes required in MadMatrix are those in [madgraph4gpu#1086](https://github.com/madgraph5/madgraph4gpu/pull/1086).

Also, let's not forget [madgraph4gpu#1090](https://github.com/madgraph5/madgraph4gpu/pull/1090), for the related changes in MadtRex.

PR #76, opened 2026-08-19
Backend separation

stloufra:
## Summary

Physically separates backend-owned driver code (cpu / simd / gpu) out of the madmatrix generator's shared and per-process-generated files into `backend/{cpu,simd,gpu}/`, so a given backend's code lives in its own directory instead of being selected at compile time via `#ifdef MGONGPUCPP_GPUIMPL` / `MGONGPU_CPPSIMD`.

## What changed

- **New layout**: `SigmaKin`, `color_sum`, `CrossSectionKernels`, `MatrixElementKernels`, `EventStatistics`, `umami`, the 14 `MemoryAccess*.h` files, and `mgOnGpuConfig/Cxtypes/Fptypes/Vectors.h` now live under `backend/{cpu,simd,gpu}/`, sourced by the Makefile via `-I`/`vpath` instead of being copied per process.
- **New per-process files**: `ColorData.h` (merges colour channel information + the color matrix), `ProcessData.h` (merges contexpr process data e.g. `nwf`, extended with the helicity/flavor tables that used to be inline in `CPPProcess.cc`), `ProcessTables.h` (flavor-coupling / broken-symmetry tables) — single source of truth read by both `CPPProcess` and backend-owned code.
- **Shrunk `CPPProcess.cc`**: all funcs related to `SigmaKin` computation moved to `/backend/Sigmakin` `fpeEnable()`/`getCompiler()` moved to backend files; per-parameter imperative literal-assignment code replaced by a generated pointer-to-member table plus one generic `gatherFptype`/`gatherCxtype` loop written once in `Parameters.h`.
- **Deleted dead code**: `Bridge.h`, `BridgeKernels.*`, `fbridge.*`, `fsampler.cc` (zero remaining references); 
- **`GpuAbstraction.h`/`GpuRuntime.h` made gpu-only.
- **Namespace unification**: `mg5amcCpu`/`mg5amcGpu` → one `madmatrix` namespace (the split is now physical, not namespace-based)
- **Backend call sites, no more `#ifdef` at the point of use**:
  - `createNormalizedColorMatrix()`: no-op inline on cpu/simd, real `gpuMemcpyToSymbol` on gpu.
  - `ColorData.h`'s device→host copy of `channel2iconfig` removed entirely (its only consumer, `runTest.cc`, isn't wired into any export path — kept a local `#ifndef`/`#else` there since it can't get per-process `%`-substituted literals the way `ColorData.h` can for now).
- **Removed `backend/{cpu,gpu}/mgOnGpuVectors.h` entirely** — neither backend has real C++ vector types; their scalar-only content (`neppV`, `fpternary`/`cxternary`/`maskand`, the `_sv` typedefs, `cxabs2`) merged into `mgOnGpuCxtypes.h`. The handful of shared, non-per-backend `.inc` files that need this now pick the right header via `__has_include`. Stripped the remaining always-true `#ifdef MGONGPU_CPPSIMD` guards (and the dead code blocks they wrapped) from `backend/simd/mgOnGpuVectors.h`.
- Standalone files are guarded for `output standalone_mg7` only
- `USEBUILDDIR` defaults to 1 for cleaner builds
- `rambo` used in standalone mode for both massive and massless version as well as the random number kernels live in `./src/rambo` and is sourced when `P1_*` is compiled 

## 

Net effect
- defined `MGONGPUCPP_GPUIMPL` now survives only for poping diagnostics while carrying `hip/cuda` diversification and in the standalone specific files (`rambo/check_sa`)
- defined `MGONGPU_CPPSIMD` now survives only in the /backend/simd as carry for the sims lane size and in the `HelAmps_sm.h` polarisation functions for masking paths of conditional code 

## Test

- [x] Full 4-backend (`cppnone`/`cppsse4`/`cppavx2`/`cuda`) × 3-process (`gg_ttxgg`, `uu_huugg`,`epem_mupmum`) bitwise match  on our a100 machine (GCC + GPU) and Apple M4 Pro (Clang) checking 64 `MatrixElement` and `MeanMatrixElemValue` against the pre-refactor baseline commit (`1a4f00df67`, the brach from commit on `main`) using `check_sa -p -v 256 64 1`

## Refs 
- Summary presentation
[backend_separation.pdf](https://github.com/user-attachments/files/31224987/backend_separation.pdf)
- Script for performing previous check 
[check_backend_regression.py](https://github.com/user-attachments/files/31225173/check_backend_regression.py)


theoheimel:
There was a bug on MacOS where the absolute path of the common library was hardcoded into the .so, preventing the user to move runs around. I've applied a temporary fix, https://github.com/MadGraphTeam/MadGraph7/pull/97 . We'll need to reapply this here

stloufra:
@Theo thanks for heads up. Should be resolved when both merged into main or should I cherry pick some specific commit?

theoheimel:
I don't know how much you have refactored the whole makefile business. I guess it will show up as a merge conflict. In any case, this was the commit: [35af7da](https://github.com/MadGraphTeam/MadGraph7/pull/97/commits/35af7daa7ce2c5e945bac66ad6614c25cc4db6ea)

Qubitol:
Superseeded by #166.

PR #77, opened 2026-08-20
Move madspace code generation out of the source tree

theoheimel:
The madspace source contains some generated code that often leads to large diffs and merge conflicts. This runs the codegen incrementally at build time and removes these files from the source tree.


oliviermattelaer:
Nice, let's merge

PR #80, opened 2026-08-21
loop-induced: stop the run when the poles do not cancel

oliviermattelaer:
A loop-induced amplitude is finite: its single- and double-pole coefficients must
vanish. Nothing checked that. This adds a check that stops the run when they do not.

## Nothing was checking the poles

Three things looked like a pole check and none was one:

- **`check_poles` never reaches a loop-induced process.** It is linked only by the FKS
  exporter; `[noborn=]` is exported through the LO/madevent templates, which have no
  `check_poles.f` and no `BinothLHA.f`. It is a different test in any case — it compares
  MadLoop's poles to the MadFKS Born-proportional ones.
- **The `##W04`/`##W05` blocks** in `madgraph/loop/loop_exporters.py` are emitted as
  Fortran comments AND live only in the non-optimized exporter. The optimized exporter —
  the default, and what every loop-induced madevent run uses — had no pole test at all.
- **A runtime test at `matrix_loop_induced_madevent.inc:418`** is dead three ways:
  `WRITE`-only, capped at 10 points, and it sums the *signed* `OneEps+TwoEps`, so a
  positive single pole can cancel a negative double pole and report clean.

## The check

New `MLPoleCheckThres` MadLoopParams option, default `1.0d-3`, negative to disable,
emitted inside the generated MadLoop matrix element and gated on loop-induced at
generation time. That is the one place with access to every guard, and it covers
madevent, `check_sa`, f2py and OLP at once.

Also fixed along the way: the `##W04` block tested each loop amplitude *individually*,
which is wrong physics — individual loop diagrams do have poles, only the sum is finite,
and it compared them against a squared-ME scale, so it was dimensionally wrong too. The
non-optimized loop-induced squaring was linear in the pole while `ANS(1)` is quadratic;
the historical `10**5` fudge existed only to paper over that, so the squaring is fixed
rather than the fudge inherited. The signed sum in the madevent driver becomes
`ABS(OneEps)+ABS(TwoEps)`. The standalone exporter no longer disables COLLIER's pole
computation, otherwise the check could never run there.

## The threshold comes from measurement

1048 physics-phase points over four processes, check disabled so nothing truncates the
sample:

| process | median | max |
|---|---|---|
| `g g > z z` | 8.6e-14 | 5.9e-10 |
| `g g > h h` | 1.3e-14 | 3.9e-09 |
| `g g > z z g` | 2.1e-11 | 4.0e-05 |
| `g g > h h g` | 4.7e-13 | 5.6e-05 |

Ceiling 6e-5. `1e-4` was rejected at 1.8x headroom; `1e-3` gives ~20x and detects errors
of 0.7% or more, against 6% at the old `1e-2`, with no additional false positive on this
sample. The 2->2 / 2->3 gap is a singular region *appearing*, not conditioning degrading
with leg count, so a 2->4 process adds more regions of the same character rather than
another four orders.

## Evidence it is not vacuous

Injecting a 10% error on the eps^-1 residue of a **dominant** loop amplitude gives
`##E03` and exit 1. End to end at the shipped default: clean run exit 0 (median 1.69e-14),
0.5% error not caught, 1% caught, 10% caught. Verified independently by a second reviewer,
who also confirmed the corrupted value genuinely travels through
`COMPUTE_COLOR_FLOWS -> JAMPL -> ANS(2,0)` rather than being fed back into the test.

## Limitations, stated in the card rather than omitted

- **Points on a soft or collinear singularity can legitimately reach 5e-2.** madevent
  never sees them (cuts run before the matrix element), but `check_sa` and `launch` feed
  uncut phase-space points. The card says so and says to raise the value there. Note this
  is not new to the threshold change: such a point exceeds the old `1e-2` too. The durable
  fix is to warn for the first N violations and stop only on a repeat — recorded as a
  follow-up, not implemented here.
- **MadLoop's own stability verdict does not bound the pole residue**: the outlier above
  returns `rc=217`, H=2, "stable, no rescue needed".
- **An error on a subdominant amplitude is not caught at any usable threshold.**
- **The madevent event loop with COLLIER installed stays inert by design** (poles are
  turned off there for speed and the card cannot override it). The check is live in
  standalone, `check_sa`, `launch`, on any install without COLLIER, and in madevent when
  `nhel != 0`. The `##INFO` announcing the skip no longer offers a remedy that cannot work.
- Re-enabling COLLIER's poles in standalone costs ~2.5x per point (marginal, N=1000).

## Non-loop-induced output is untouched

`g g > t t~ [virt=QCD]` with `PYTHONHASHSEED=0`: ten generated files md5-identical
base vs branch. IOTest goldens added for the optimized loop-induced path, which had none.


oliviermattelaer:
Closing — dropped on Olivier's call. The branch `claude/loop-induced-pole-check` is left in place, so nothing here is lost if it is wanted later.

What it contained, for anyone reading back:

- **The finding stands regardless of this PR: nothing checks that loop-induced poles vanish.** `check_poles` never reaches a loop-induced process (it is linked only by the FKS exporter; `[noborn=]` goes out through the LO/madevent templates, which have no `check_poles.f` or `BinothLHA.f`). The `##W04`/`##W05` blocks in `loop_exporters.py` are emitted as Fortran comments *and* live only in the non-optimized exporter, so the optimized one — the default, used by every loop-induced madevent run — had no pole test at all. A third runtime test at `matrix_loop_induced_madevent.inc:418` is dead three ways: WRITE-only, capped at 10 points, and it sums the *signed* `OneEps+TwoEps`, so a positive single pole can cancel a negative double pole and report clean.
- A `MLPoleCheckThres` option, default `1.0d-3`, set from 1048 measured phase-space points across four processes (median 1.3e-14 to 2.1e-11; ceiling 6.0e-5 on `g g > h h g`). `1e-4` was rejected at 1.8x headroom.
- Two limitations the card documented: points on a soft or collinear singularity can legitimately reach 5e-2 (madevent never sees them, but `check_sa` and `launch` feed uncut points), and MadLoop's own stability verdict does not bound the pole residue — the outlier returns `rc=217`, "stable, no rescue needed".

Also fixed on the branch, and now unlanded with it: the `##W04` block tested each loop amplitude *individually*, which is wrong physics — individual loop diagrams do have poles, only the sum is finite — and compared them against a squared-ME scale, so it was dimensionally wrong too. The non-optimized loop-induced squaring was linear in the pole while `ANS(1)` is quadratic.

PR #89, opened 2026-09-02
madspace: compute scale/PDF systematics when writing the events

oliviermattelaer:
## Summary

madspace now computes the scale and PDF variation weights itself when it writes the events (combine step), from the reweighting inputs the integrand already stores for accepted events. The `systematics.py` + LHAPDF post-processing step is no longer needed for the mg7 output; it stays available as a legacy option (`[postprocessing] systematics = true`, default off).

- **All output formats**: LHE gets an `<initrwgt>` header and `<rwgt>` blocks in the `systematics.py` conventions (plus an optional `<mgrwt>` block with `write_inputs = true`); `lhe_npy`/`compact_npy` get `rwgt_<id>` columns and an `events.weights.json` sidecar; gridpacks save and reload the configuration.
- **Variations**: mu_R x mu_F grid, alternative dynamical scale choices (recomputed from the stored momenta), PDF members of the nominal set (`errorset`) and of other sets (with their own alpha_s), `set@member` entries.
- **Mixed QCD/QED orders**: subprocesses whose |M|^2 mixes several alpha_s powers get exact mu_R weights by re-evaluating the matrix element at the varied alpha_s on a dedicated CPU context (`qcd_power` is now recorded per subprocess in `subprocesses.json`).
- **Histograms with bands**: the `[histograms]` observables are also filled on the written sample with every variation weight; `info.json["event_histograms"]` carries per bin the nominal value, binomial error, scale envelope and PDF uncertainty.
- **Summary**: per-variation cross sections, scale envelope and PDF uncertainty (by `ErrorType`) in `info.json["systematics"]` and in the log.

New run-card section:

```toml
[systematics]
enable = true
mur = [0.5, 1.0, 2.0]
muf = [0.5, 1.0, 2.0]
together = true
dynamical_scale = []
pdf = ["errorset"]
write_inputs = false
```

`[generation] systematics` is kept as a hidden alias of `enable`. With a one-member nominal set (e.g. NNPDF40MC_lo_as_01180) `errorset` yields no PDF weights and a warning naming the set.

Design and full validation record: `docs/madspace-systematics-plan.md` (Section 7).

## Validation

- Scalar PDF / alpha_s interpolation vs the runtime kernels: identical (max rel. diff 0.0); vs LHAPDF 1e-15 / 3e-16.
- Native weights vs `systematics.py` + LHAPDF, weight by weight, same file: < 1e-7 (the LHE 7-digit format) for `u u > u u` (8 scale + 101 PDF weights), `p p > e+ e-` (mirrored flavours), `u g > u g`, and with the 4 dynamical scale choices (37 weights).
- Matrix-element reweighting forced on the pure-QCD `u u > u u` vs the analytic alpha_s^2 rescaling: 2e-7.
- Event histograms: per-variation bin sums equal the per-variation cross sections; envelope brackets the nominal; shapes agree with the integration histograms.
- `lhe`, `lhe_npy`, `compact_npy` and gridpack runs produce the same weights; the mixed-order gridpack reloads its matrix element.
- Tests: `madspace/tests/test_systematics.py` (11), the madspace suite, and the acceptance tests `test_systematics_mg7`, `test_systematics_mg7_legacy`, `test_systematics_mixed_order_mg7` (need LHAPDF data).

## Speed

`u u > u u`, 3x3 scales + 101 NNPDF2.3 replicas (109 weights/event), same file for both tools:

| Events | Native combine (wall / CPU) | Whole run with / without systematics | Legacy `systematics.py` (wall / CPU) |
|---|---|---|---|
| 20 000 | 0.28 s / 3.5 s | 2.1 s / 0.3 s | 21.6 s / 21 s |
| 100 000 | 1.3 s / 21 s | 2.6 s / 1.0 s | 1025 s / 100 s |

About 1 s of the native overhead is parsing the 100 member grids, independent of the event count.

## Not done

- Physics closure (direct runs at the varied scale/PDF vs the reweighted cross sections).
- Pythia8 / MadSpin not re-run on the new files (header and per-event blocks follow the existing `systematics.py` conventions).
- The full MadGraph unit suite was not re-run; the run-card unit tests pass.
- The integration-time `histograms` block in `info.json` is not normalised to the cross section (pre-existing); the event histograms are.


oliviermattelaer:
Thanks, agreed. Done in 04805cbd9: the scalar `PdfGrid::interpolate` / `AlphaSGrid::interpolate` path is removed (`pdf.hpp`/`pdf.cpp` are back to their original content). The calculator now registers the nominal grid and every varied member as globals of a CPU context (private prefix, restricted to the PIDs the events use) and evaluates the regular batched `PartonDensity` / `RunningCoupling` functions through runtimes, one call per grid and one per distinct alpha_s point for a whole combine batch. The matrix elements of the mixed-order subprocesses live on the same context, so an UMAMI-based PDF would plug in at the same place.

Validation is unchanged (native vs `systematics.py` + LHAPDF < 1e-7, ME path vs analytic rescaling 2e-7, acceptance tests green) and it is faster: 100k events x 109 weights combine in 0.9 s wall / 2.8 s CPU (was 1.3 s / 21 s). The one trade-off is memory: the coefficient tensors of the varied members stay resident, 0.66 MB per PID per member (66 MB for 100 replicas of a single-flavour process, ~0.7 GB with all 11 light partons).

oliviermattelaer:
@theoheimel in principle the CI should be green now (but we have again a cache issue, and the cache is in reconstruction for the moment)

oliviermattelaer:
Todo before merging
- investigate the CI crash
- change default PDF

theoheimel (approved):
Approved once CI passes

PR #97, opened 2026-09-05
Remove absolute paths to allow moving runs and gridpacks

theoheimel:
- use relative path in `subprocesses.json`
- on MacOS: use relative path for linking to the common .so file in madmatrix

### DIFF
--- a/madgraph/iolibs/template_files/madmatrix/madmatrix.mk
+++ b/madgraph/iolibs/template_files/madmatrix/madmatrix.mk
@@ -116,6 +116,9 @@ override SRC := ../../src
 # Can be overridden on the command line (e.g. make LIBDIR=/abs/path); defaults to ../../lib.
 # NB: LIBDIR is resolved to an absolute path so it can be passed unchanged to sub-makes via export.
 LIBDIR ?= ../../lib
+# Keep LIBDIR as given (usually relative to this P* directory): it is used below to
+# build loader-relative rpaths, which survive a move of the process directory.
+override LIBDIR_AS_GIVEN := $(LIBDIR)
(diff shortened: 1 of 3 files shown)


madgraph/iolibs/template_files/madmatrix/madmatrix.mk
```diff
@@ -116,6 +116,9 @@ override SRC := ../../src
 # Can be overridden on the command line (e.g. make LIBDIR=/abs/path); defaults to ../../lib.
 # NB: LIBDIR is resolved to an absolute path so it can be passed unchanged to sub-makes via export.
 LIBDIR ?= ../../lib
+# Keep LIBDIR as given (usually relative to this P* directory): it is used below to
+# build loader-relative rpaths, which survive a move of the process directory.
+override LIBDIR_AS_GIVEN := $(LIBDIR)
```
Qubitol, line 121: I'm a bit puzzled, why having also `LIBDIR_AS_GIVEN`? Can't we just remove the `abspath` from `LIBDIR`? And then add it again later in the if statement?

theoheimel: This was done by Claude. If the fix is overcomplicated (as AI fixes tend to be), feel free to replace it with something simpler. I don't really know the madmatrix makefile logistics, so the `LIBDIR_AS_GIVEN` might not be necessary. I guess, this will be replaced after the backend separation merge anyways.

Qubitol: Ah yes, at the end we are gonna change it afterwards anyway, so it doesn't make sense to do it right now.

PR #104, opened 2026-09-07
Release workflow: publish a GitHub release to trigger tests, PyPI and the tarball (#78)

### LINKED ISSUE #78: Set up release workflow
Set up a release workflow that triggers automatically when a release is added on GitHub.

Requirements:
- run full test suite and require it to pass
- generate the tarball for madgraph and add it to the release
- trigger a madspace release on PyPI with the same version number as the MG release
- change the madspace installer script such that it offers to use the binary madspace release (only in releases and with the exact same version as MG) instead of compiling from source

theoheimel:
Closes #78.

Adds `.github/workflows/release.yml`, triggered by `release: published`,
which:
1. checks the release tag agrees with `VERSION` and `madspace/pyproject.toml`,
2. warms the HEPTools/UFO/pip caches (`restore_heptools` hard-fails on a cache
   miss, so a release can't quietly depend on the cache having stayed warm),
3. runs the full test suite (unit, acceptance, madevent, mg7, parallel, IOTests,
   aloha, madspin, xsec) by making those workflows `workflow_call`-able,
4. builds the full madspace wheel matrix and publishes to PyPI (madspace is
   never released standalone — its version always matches the MadGraph
   release),
5. builds the MadGraph tarball (`bin/create_release.py`, rewritten from the
   old bzr-era script), smoke-tests it (VERSION symlink survives packaging,
   `madspace/install.py --bin` picks up the wheel just published in step 4),
   signs it with Sigstore, and attaches it to the release,
6. reverts the release to draft if anything above fails.

Also: `madspace/install.py` now offers/pins the pre-compiled PyPI wheel only
in an actual release tarball (`input/.release` marker written by
`create_release.py`), and `VERSION` / `madspace/pyproject.toml` /
`docs/source/conf.py` are bumped to `0.2.0` for the alpha (both packages
share one version; avoids clashing with the existing `madspace` 0.1.x PyPI
releases).

- [x] **Manual follow-up, can't be done from the repo:** point the `madspace` PyPI
trusted publisher at workflow `release.yml` (repo `MadGraphTeam/MadGraph7`,
environment `pypi`) before the first real release.


oliviermattelaer:
I guess we need to discuss/agree on some point here:

  1. We still need something like UpdateNotes.txt (but maybe better to move to a .md extension/format and start clean)
  4. Materialize the default config/run-card files.
-> run-card should be default per process not per release
  6. Best-effort: vendor offline copies of collier/ninja/SMWidth/the
     HEPToolsInstaller bundle into vendor/.
-> This should be a blocker if missing

Otherwise the pypi stuff is something we will have to test/adapt. I do not have any clear view/experience on that

theoheimel:
The PyPI release workflow is already in use to release madspace, so that part should work. Regarding the other points, feel free to add them here. Update notes are of course something we will have to do by hand, so that's independent of this PR.

Qubitol:
> 3\. Best-effort: vendor offline copies of collier/ninja/SMWidth/the
>         HEPToolsInstaller bundle into vendor/.

I know many people don't like them, but have we ever considered using submodules to pin precise versions of the tools?
Then, when cloning one can also clone the tools using `git submodule update --init` and the submodule update is just a change in the pinned version.

I know they are not super popular, but they work pretty well if you don't plan to change them often.

oliviermattelaer:
Given that we are not authors of those tools and that some are not using git /... 
I found that complicated. I think the way is done here is fine (we shipped the currently available/tested one)
and allow user to install that one, or the currently available one.

If a user use git, then they are no pinned version at all and should install the latest one.

theoheimel:
> 2\. Materialize the default config/run-card files.
>         -> run-card should be default per process not per release

This part is basically the same as before.
Old create_release.py:
```python
# 1. Copy the .mg5_configuration_default.txt to it's default path
shutil.copy(path.join(filepath, 'input','.mg5_configuration_default.txt'), 
            path.join(filepath, 'input','mg5_configuration.txt'))
if os.path.exists(path.join(filepath, 'input','.default_run_card_lo.dat')):
    shutil.copy(path.join(filepath, 'input','.default_run_card_lo.dat'),
            path.join(filepath, 'input','default_run_card_lo.dat'))
    shutil.copy(path.join(filepath, 'input','.default_run_card_nlo.dat'),
            path.join(filepath, 'input','default_run_card_nlo.dat'))
shutil.copy(path.join(filepath, 'input','proc_card_default.dat'), 
            path.join(filepath, 'proc_card.dat'))
```
New:
```python
def materialize_config(filepath):
    input_dir = pjoin(filepath, 'input')
    shutil.copy(pjoin(input_dir, '.mg5_configuration_default.txt'),
                pjoin(input_dir, 'mg5_configuration.txt'))
    for card in ('default_run_card_lo.dat', 'default_run_card_nlo.dat',
                 'default_run_card_mg7.toml'):
        src = pjoin(input_dir, f'.{card}')
        if path.exists(src):
            shutil.copy(src, pjoin(input_dir, card))
    shutil.copy(pjoin(input_dir, 'proc_card_default.dat'),
                pjoin(filepath, 'proc_card.dat'))
```
So this behavior has always been there, in a non-process-specific way. Not sure about the purpose. Do we want to keep it or drop it?

oliviermattelaer:
Let's merge this as soon as the CI is green!!! (I have re-submitted everything that was red)

PR #105, opened 2026-09-07
Fix the PDF path logic (#94)

### LINKED ISSUE #94: Bugs in PDF path logic
There are at least two bugs in the logic for locating the PDF data path:
- PDFs are downloaded into the HEPTools folder of a different MadGraph installation. In some cases this has led to crashes. This is caused by some auto-generated config-files under `~/.mg5` that are shared between MG installations (and even shared between MG5 and MG7, which should not happen). Indeed, the issue goes away after running `rm -rf ~/.mg5`as a workaround
- When re-launching the same process through `bin/generate_events` instead of the MG CLI, I get the following error:
  ```
  RuntimeError: Can't load lhapdf module. Please set LHAPDF_DATA_PATH manually
  ```

theoheimel:
Fixes #94.

## The two bugs

**`bin/generate_events` could not find LHAPDF on its own.**

```
RuntimeError: Can't load lhapdf module. Please set LHAPDF_DATA_PATH manually
```

All the discovery lived in the caller: `do_launch` probed the configured `lhapdf-config` and injected `LHAPDF_DATA_PATH` / `MADGRAPH_LHAPDF_CONFIG` into the subprocess environment, and its final fallback *always* set the former — which is exactly why a CLI `launch` worked and re-launching the same directory did not. The standalone script sets no environment, `Cards/me5_configuration.txt` recorded only `mg5_path`, so `launch.py` was left with a bare `import lhapdf`.

**PDF sets were downloaded into another installation's `HEPTools`.**

`heptools_install_dir` defaults to the truthy `'./HEPTools'`, so `advanced_install` always took the branch that saves tool paths into the *per-user* config — as absolute paths inside whichever installation ran `install`. Every other installation read them back and, since the paths exist on disk, kept them. Hence `rm -rf ~/.mg5` working around it.

## What changed

- **One resolver.** `misc.resolve_lhapdf` is now the single place LHAPDF is located, used by both `do_launch` and the standalone launcher, so they cannot disagree. It returns the `lhapdf-config`, an ordered list of PDF-set directories, and the one a missing set may be downloaded into. It splits path lists on `os.pathsep` (feeding a multi-entry `LHAPDF_DATA_PATH` to `os.path.join` silently produced a bogus path), roots a relative option value at the installation rather than at cwd, and drops a `--python=` filter.
- **`launch.py` resolves its own paths** and exports what it finds, so systematics and the MadSpin/reweight subprocesses keep reading the environment. Side effect: the LHE banner's `lhaid` is now correct for a standalone run instead of `0`. The error message names the option and the file to edit.
- **A process directory records what it was built against.** The mg7 exporter writes `lhapdf`/`lhapdf_py3`/`heptools_install_dir` into `Cards/me5_configuration.txt`, and `load_mg5_options` reads that card *last* — process directory, then installation, then per-user, matching `CommonRunCmd.set_configuration`. Previously the per-user file overrode the installation's own.
- **`install <tool>` records paths per installation** unless `heptools_install_dir` points somewhere genuinely shared, and `Cmd.do_save` prefers a process directory's card over the per-user file.
- **MG7 keeps its configuration to itself.** The per-user location is `$XDG_CONFIG_HOME/mg7` or `~/.mg7`; MG5aMC's `~/.mg5` is never read or written. The first commit renames the in-tree file to `input/mg7_configuration.txt` to match.

## Commits

1. `chore:` rename the installation config to `mg7_configuration.txt` — mechanical.
2. `fix:` the PDF path logic.

## Verification

- 19 new unit tests in `tests/unit_tests/various/test_lhapdf_resolve.py` cover the resolver (relative values, `--python=` suffix, colon-separated `--datadir`, LHAPDF 5 vs 6, `lhapdf_py3` fallback, a broken or absent `lhapdf-config`, read-only data dir, env precedence) and the per-user location. Plus the `install` target choice in `tests/unit_tests/interface/test_cmd.py`.
- End to end with **no** `LHAPDF_DATA_PATH` and no `MADGRAPH_LHAPDF_CONFIG` in the environment: `output` a `p p > e+ e-` directory, then `cd` into it and run `./bin/generate_events -f` — exits 0, systematics included, and the banner carries `lhaid = 338500`. The CLI `launch` path was re-checked and still works.
- The manual `env['LHAPDF_DATA_PATH']` workarounds in `test_check_xsec_processes_mg7.py` and `madevent_comparator.py` are removed, so those tests now exercise the real config-driven path; the LHAPDF probe there is reduced to a skip gate.

## Note for reviewers

Anyone keeping global settings in `~/.mg5/mg5_configuration.txt` (or `~/.config/mg5_configuration.txt`) will find MG7 no longer reads them — that is the intended separation. There is no compatibility shim: at alpha stage an installation just re-seeds `input/mg7_configuration.txt` from the template on first run.

### DIFF
--- a/.github/actions/checkout_mg5/action.yml
+++ b/.github/actions/checkout_mg5/action.yml
@@ -7,7 +7,7 @@ runs:
     # setup other environment
     - run: |
         cd $GITHUB_WORKSPACE
-        cp input/.mg5_configuration_default.txt input/mg5_configuration.txt
+        cp input/.mg7_configuration_default.txt input/mg7_configuration.txt
         cp Template/LO/Source/.make_opts Template/LO/Source/make_opts
       shell: bash
 
--- a/.github/actions/restore-pip-cache/action.yml
+++ b/.github/actions/restore-pip-cache/action.yml
@@ -27,9 +27,9 @@ runs:
         # Use direct path because GITHUB_PATH additions don't take effect in the same step
         F2PY_BIN="$HOME/.cache/pip-packages/bin/f2py"
         if [ -x "$F2PY_BIN" ]; then
-          echo "f2py_compiler = $F2PY_BIN" >> input/mg5_configuration.txt
+          echo "f2py_compiler = $F2PY_BIN" >> input/mg7_configuration.txt
         else
           pip install --target "$HOME/.cache/pip-packages" --upgrade numpy
-          echo "f2py_compiler = $F2PY_BIN" >> input/mg5_configuration.txt
+          echo "f2py_compiler = $F2PY_BIN" >> input/mg7_configuration.txt
         fi
       shell: bash
--- a/.github/actions/restore_delphes/action.yml
+++ b/.github/actions/restore_delphes/action.yml
@@ -32,5 +32,5 @@ runs:
 
     - run: |
         cd $GITHUB_WORKSPACE
-        echo "delphes_path = $HOME/.cache/Delphes/Delphes" >> input/mg5_configuration.txt
+        echo "delphes_path = $HOME/.cache/Delphes/Delphes" >> input/mg7_configuration.txt
       shell: bash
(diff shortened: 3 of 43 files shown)


oliviermattelaer (approved):
We can merge this. Thanks a lot,

PR #108, opened 2026-09-08
Mixed precision for all backend

stloufra:
`FPTYPE=v` (mixed precision)  - momenta + propagator denominator in FP64, wavefunctions/amps/colour in FP32.
For SIMDS introduction of pair types (`fptype_denom_sv/cxtype_denom_sv` = two native FP64 vectors): every operator touches both halves, so the denominator is evaluated twice at native width to fill the FP32 lane count.
Generated HelAmps_sm.h denom line is unchanged.
Non-v builds untouched: `fpamp_of_mom/fpamp_scalar` are macros - narrowing/casting.
f/d bitwise-identical to main on all 5 backends.
v vs d: ≤ 1.3e-5 relative error (limit 1e-3). Compiles & runs for 9 processes (g g > t t~  + ng, u. u~ > e+ e- +ng, u u  > h u u g g)  on V100S/A100/L40S and all supported vectorisations (Intel(R) Xeon(R) Silver 4216).
Throughput v/d: ~1.5–1.9× on CPU SIMD; GPU launch-bound for tiny processes, up to ~2–3× for amplitude-heavy ones (L40S gg→tt̄ggg = 2.95×).
New code `#ifdef MGONGPU_SIMD_DENOM64`-guarded.


Qubitol:
Fantastic. Shall we have the mixed precision `FPTYPE=v` as default?

stloufra:
After merging main some numerical issues are showing up. Need to understand before merge.

stloufra:
Reopening the PR - bug fixed. Also enclosing a summary pdf for the throughput enhancement as well as accuracy tests. with the [latest](https://github.com/MadGraphTeam/MadGraph7/pull/108/commits/38eec445ce43da58ecae1ffc4a82079593dc9024).

[PR_mixed.pdf](https://github.com/user-attachments/files/32008353/PR_mixed.pdf)

Back compatibility checked `f` and `d` for 40000 points.

stloufra:
Before merge, CI should be green - looking into that.

oliviermattelaer:
I added a "check precision MODE PROCESS --nb_event="
example:
```
FPTYPE=f vs FPTYPE=d, 1000000 events per flavour
  f = all32: single precision everywhere
Process        mode  mean error      max error       rate > 1%       non-finite  time d [s]      time mode [s]   speed-up
g g > g g g g  f     1.597e-05       1.445e+00       1.060e-04       0           1.973e+02       1.004e+02       1.97
(time: matrix element evaluation only, for the 1000000 events)
Plot of the difference: /Users/omattelaer/Documents/git_workspace/MadGraph7/.claude/worktrees/merge-main-resolve-conflicts-2ab60b/check_precision_f_P0_gg_gggg.pdf
```

we can check more than one mode:
check precision m v g g >4g
```
Process        mode  mean error      max error       rate > 1%       non-finite  time d [s]      time mode [s]   speed-up
g g > g g g g  m     4.549e-08       3.213e-07       0.000e+00       0           1.961e+02       1.843e+02       1.06
g g > g g g g  v     4.602e-07       1.075e-04       0.000e+00       0           1.961e+02       1.172e+02       1.67
(time: matrix element evaluation only, for the 1000000 events)
Plot of the difference: /Users/omattelaer/Documents/git_workspace/MadGraph7/.claude/worktrees/merge-main-resolve-conflicts-2ab60b/check_precision_m_v_P0_gg_gggg.pdf
```

oliviermattelaer:
So as soon as the CI is green we can merge this. Well done.

oliviermattelaer (approved):

oliviermattelaer:
Screenshot 2026-09-13 at 17 46 55

Quite impressive !!

PS: One need to investigate issue with the axial gauge. But we can merge this first and investigate independently

roiser (commented):
Very fine with me, just a cosmetic change, I see there are new files, please put the usual heading comment block with copyright and license on those. Also feel free to add yourself in the he list of authors where you changed files. After that please feel free to merge from my side. 

Well done !!!

madgraph/iolibs/template_files/madmatrix/Arithmetics/basicOPs.h
```diff
@@ -0,0 +1,125 @@
...
+         std::is_same_v<T, double_st> || std::is_same_v<T, float_st>;
+#endif
+
+#ifdef __CADNA__
+#define FLOAT_TEMPLATE_GUARD \
+template< typename T, std::enable_if_t< is_special_fp_v< T >, int > = 0 >
+#else
+#define FLOAT_TEMPLATE_GUARD \
+template< typename T, std::enable_if_t< std::is_floating_point_v< T >, int > = 0 >
+#endif
```
Qubitol: Given our love-hate relationship with macros, I'd suggest using a concept. MadGraph7 requires C++20, so concepts are supported:
```suggestion
#ifdef __CADNA__
template <typename T>
concept GoodFloatType = std::is_same_as<double_st, T> || std::is_same_as<float_st, T>;
#else
template <typename T>
concept GoodFloatType = std::is_same_as<double, T> || std::is_same_as<float, T>;
#endif
```
Also, while trying to understand `is_floating_point_v`, I found out that it can accept also `long double`, which maybe we don't want, so it's good to list explicitly double and float in the concept defintion in my opinion (and it makes a nice symmetry with the CADNA case).
*Also, the name `GoodFloatType` is pretty terrible, so feel free to change it*

madgraph/iolibs/template_files/madmatrix/Arithmetics/basicOPs.h
```diff
@@ -0,0 +1,125 @@
...
+
+
+FLOAT_TEMPLATE_GUARD
```
Qubitol: On the same page with concepts, we would have
```suggestion
template <GoodFloatType T>
```
here.

madgraph/iolibs/template_files/madmatrix/Arithmetics/basicOPs.h
```diff
@@ -0,0 +1,125 @@
...
+static constexpr __tnl_inline__ T
+add_rn( const T x, const T y )
+{
+#if defined __CUDA_ARCH__
+   if constexpr( std::is_same_v< T, double > ) {
+      return __dadd_rn( x, y );
+   }
+   else if constexpr( std::is_same_v< T, float > ) {
+      return __fadd_rn( x, y );
+   }
```
Qubitol, line 64: I noticed that this would work only in case of `T` being `double` or `float`.
What about the CADNA types?
Does it make sense to have a single CADNA macro at the beginning that defines `dtype` and `ftype` to either `double`/`float` or `double_st`/`float_st` and then we keep using `dtype` and `ftype` so that it works without having many CADNA macro scattered around (even in the concept case)?

madgraph/iolibs/template_files/madmatrix/Arithmetics/basicOPs.h
```diff
@@ -0,0 +1,125 @@
+#ifndef BASICOPS_H
+#define BASICOPS_H
+
+#pragma once
```
Qubitol: If we rely already on `#ifndef` directive, then I guess specifying also `#pragma once` is redundant.

madgraph/iolibs/template_files/madmatrix/Arithmetics/Double.h
```diff
@@ -0,0 +1,413 @@
...
+  __cuda_callable__
+  constexpr Double() = default;
+
+  __cuda_callable__
+  constexpr Double( const Double& other ) = default;
+
+  __cuda_callable__
+  constexpr Double( Double&& other ) noexcept = default;
+
+#ifdef __CADNA__
```
Qubitol: See suggestions above to consider the usage of concepts.

madgraph/iolibs/template_files/madmatrix/Arithmetics/Double.h
```diff
@@ -0,0 +1,413 @@
...
+   constexpr Double< T >
+   operator*( const Double< T >& A, const Double< T >& B )
+  {
+     auto tp = two_prod( A[ 0 ], B[ 0 ] );
+     /*volatile*/ const T temp = add_rn( mul_rn( A[ 1 ], B[ 0 ] ), mul_rn( A[ 0 ], B[ 1 ] ) );
+     auto qtsRes = quick_two_sum( tp.sum, add_rn( tp.error, temp ) );
+     return Double< T >( qtsRes.sum, qtsRes.error );
+  }
+
+/*template< typename T, typename U, std::enable_if_t< std::is_arithmetic_v< U > , int > >
```
Qubitol: Is this huge commented part of the code necessary?

madgraph/iolibs/template_files/madmatrix/Arithmetics/errorFreeOPs.h
Qubitol, line 1: Same comments about considering concepts, and about understanding whether the currently commented chunk of this file (towards the end) is necessary.

madgraph/iolibs/template_files/madmatrix/Arithmetics/errorFreeOPs.h
```diff
@@ -0,0 +1,225 @@
+
+#ifndef ERRORFREEOPS_H
+#define ERRORFREEOPS_H
+
+#pragma once
```
Qubitol: Same comment as before.

madgraph/iolibs/template_files/madmatrix/madmatrix.mk
```diff
@@ -41,7 +41,7 @@ ifneq ($(words $(filter $(BACKEND), $(SUPPORTED_BACKENDS))),1)
   $(error Invalid backend BACKEND='$(BACKEND)': supported backends are $(foreach backend,$(SUPPORTED_BACKENDS),'$(backend)'))
 endif
 
-override SUPPORTED_FPTYPES = d f m
+override SUPPORTED_FPTYPES = d f m e v
```
Qubitol, line 46: I'd suggest putting a little comment explaining what each of these letters refers to, since it might be a bit cryptic in case of `e` and `v`.
I know we will have documentation and so on, but a little comment here would not be bad.

madgraph/iolibs/template_files/madmatrix/madmatrix.mk
```diff
@@ -600,17 +600,26 @@ endif
 
 # Set the build flags appropriate to each FPTYPE choice (example: "make FPTYPE=f")
 $(info FPTYPE='$(FPTYPE)')
-ifeq ($(FPTYPE),d)
+# 3 precision macros: amp (MGONGPU_FPTYPE_*), colour (MGONGPU_FPTYPE2_*), momenta/denom (MGONGPU_FPTYPE_MOMENTA_*)
+# 4 modes: d=all64, f=all32, v=denom64, m=color32 (e for doubleword expansion)
```
Qubitol: Alright, the comment is here ~500 lines ahead, so I would not mind replicating it, or actually moving it all together at the beginning (since anyway you are spelling out the things in each `if` statement).

madgraph/iolibs/template_files/madmatrix/mgOnGpuCxtypes.h
```diff
@@ -322,6 +323,44 @@ namespace mg5amcCpu
 
   // SANITY CHECK: memory access may be based on casts of fptype[2] to cxtype (e.g. for wavefunctions)
   static_assert( sizeof( cxtype ) == mgOnGpu::nx2 * sizeof( fptype ), "sizeof(cxtype) is not 2*sizeof(fptype)" );
+
+  // --- Multi-precision complex types (same platform logic as cxtype)
+#ifdef __CUDACC__ // this must be __CUDACC__ (not MGONGPUCPP_GPUIMPL)
+#if defined MGONGPU_CUCXTYPE_THRUST
+  typedef thrust::complex<fptype_momenta> cxtype_momenta;
```
Qubitol, line 330: I wonder if it is worth keeping the CUDA types, even if I've never seen them used.

Qubitol (changes requested):
It's fantastic, I like it a lot.
First of all, the code is fine, I just made some comments to maybe consider using concepts instead of SFINAE, and to take care of an inconsistency with CADNA.

Please, check (what it seems) an inconsistency with CADNA, since if you activate CADNA with GPU now it would not pass the `is_same_v` with `double_st` and `float_st`.

madgraph/iolibs/template_files/madmatrix/Arithmetics/basicOPs.h
stloufra: I like the name :) Thanks for the concept solution.

madgraph/iolibs/template_files/madmatrix/Arithmetics/basicOPs.h
stloufra: the `__dadd_rn()` is CUDA callable func which is defined specifically for `double` FP type. In CADNA, we do not have overload for that; Also the *Double word arithmetics* is more of an experimental feature as it did not prove to be faster for tested cases on current hardware so far; I just do not want to loose the work and therefore decided to keep it for future with the PR...

Qubitol: That makes a lot of sense.
So, if we use CADNA while running on GPU with these types, it won't work?
If that is the case, we should handle that case and crash properly.

madgraph/iolibs/template_files/madmatrix/mgOnGpuCxtypes.h
stloufra: If I remember correctly the `backend_separation` is addressing the removal of the `thrust::complex<>` so I dont want to duplicate.

stloufra:
> Very fine with me, just a cosmetic change, I see there are new files, please put the usual heading comment block with copyright and license on those. Also feel free to add yourself in the he list of authors where you changed files. After that please feel free to merge from my side.
> 
> Well done !!!

Done :) thanks for the catch.

Qubitol (approved):

PR #110, opened 2026-09-08
Two-particle invariant mass cut that also bounds the integration

oliviermattelaer:
A cut on the invariant mass of a pair was only expressible as `mass` with summed momenta, which is the mass of the *whole* selection and happens to be a pair only when the selection holds exactly two particles. There was no way to cut on every pair a group can form, and no way to cut at all on a group larger than two.

More importantly, such a cut only ever filtered finished events. It never shaped the phase space, so asking for a hard cut meant generating the forbidden region and throwing nearly all of it away.

## What this adds

`obs_m_inv`, a genuine pairwise observable applied to each pair the selection produces:

```toml
lepton-m_inv.min = 500.0
top-m_inv.min    = 800.0
```

and, where the pair is exactly what a propagator decays into, the cut becomes that propagator's floor on its own invariant mass.

That last part needed almost no new machinery. Every propagator already carries `e_min`, and `update_mass_min_max` already carries it into the `s_min` handed to the invariant sampler, into what the parents subtract for their maximum, and into `s_hat_min`. Raising `e_min` is therefore enough for the associated decaying particle's invariant and sqrt(s_hat) to follow on their own.

Where no single propagator carries the pair, the cut still bounds the total invariant mass: the pair contributes at least the cut and the rest of the final state at least its masses. The smallest such bound over the named pairs is used, which is the one that holds whether the cut must be satisfied by all of them or by only one.

Both are necessary consequences of the cut, so neither removes phase space the cut would have kept.

## Measured

`p p > e+ e- j j j` with m(ee) > 200 GeV:

| | sigma [pb] | cut efficiency | RSD |
|---|---|---|---|
| before | 0.15082 +- 0.00077 | 27.8% | 41.0 |
| after | 0.15014 +- 0.00024 | 53.8% | 10.4 |

The same answer to 0.8 sigma, twice the efficiency, four times the RSD. At m(ee) > 500 the contrast is sharper: before, the survey gives up with *"not enough points passing cuts after 1000 batches"*; after, it completes at 58% efficiency.

`p p > t t~ j` with m(t t~) > 800 GeV is unchanged at 1.1 sigma, 47.714 +- 0.093 against 47.576 +- 0.084, and every generated event respects the cut, the smallest being 800.01 GeV.

## Testing

`madspace/tests/test_m_inv_cut.py`, on `g g > t t~ g g` whose first diagram has the top pair on a single propagator. It pins down that `obs_m_inv` is the pair's invariant mass and not `obs_mass`; that nothing below the cut is generated once the mapping knows about it; that without it more than half the points land where the cut would have discarded them; and that the volume of the cut region is the same either way, over a region that is a real fraction of the total.

Verified to fail without the feature: disabling the propagator floor drops the smallest generated m(t t~) from the cut to 346 GeV, twice the top mass.

Full madspace suite: no new failures against `main`. The four `test_double_t` failures on this branch are present on `main` too.

## Note for review

The second commit also fixes a bug this feature exposed: `Cuts` indexes its per-particle tables by outgoing position assuming two incoming particles, so on a 1 -> n decay topology those tables are a different length than the outgoing legs. Reading one against the other threw out of every mapping built from a decay. A decay has no cuts to apply, so the boundary is skipped there.

The gain is largest for genuinely resonance-like pairs. For `t t~ j` the pair is usually *not* an s-channel propagator in `g g -> t t~ g`, so only the sqrt(s_hat) bound applies and the gain is modest.

### DIFF
--- a/madgraph/iolibs/template_files/mg7/run_card.toml
+++ b/madgraph/iolibs/template_files/mg7/run_card.toml
@@ -90,12 +90,17 @@ $multiparticles
 
 [cuts]
 # possible groups: jet, bottom, lepton, missing, photon
-# possible observables: pt, eta, dR, mass, m_inv, sqrt_s
-#     (mass is the mass of one particle, or of the summed momenta with "-sum-";
-#      m_inv is the invariant mass of a pair, applied to every pair the group
-#      can form; sqrt_s is for all outgoing particles)
-# An m_inv minimum is also used to bound the integration: where the pair is
-# what a propagator decays into, the cut becomes that propagator's minimum
+# possible observables: pt, eta, dR, mass, pair_mass, sqrt_s
+#     (mass is the mass of one particle. Put "-sum-" before it and the momenta
+#      are added up first: "lepton-sum-mass" is the mass of all the leptons
+#      together, while naming a group per member gives one combination at a
+#      time, so "lepton-lepton-sum-mass" is the mass of a lepton pair and
+#      "lepton-jet-top-sum-mass" that of a lepton-jet-top triple.
+#      pair_mass is the invariant mass of a pair, taken over every pair the
+#      named groups can form, so "lepton-pair_mass" is the same cut as
+#      "lepton-lepton-sum-mass". sqrt_s is for all outgoing particles)
+# A pair_mass minimum is also used to bound the integration: where the pair
+# is what a propagator decays into, the cut becomes that propagator's minimum
 # invariant mass, so the forbidden region is never generated rather than
 # generated and thrown away.
 # for all cuts, min or max can be specified
--- a/madgraph/iolibs/template_files/mg7/run_card.toml
+++ b/madgraph/iolibs/template_files/mg7/run_card.toml
@@ -90,8 +90,14 @@ $multiparticles
 
 [cuts]
 # possible groups: jet, bottom, lepton, missing, photon
-# possible observables: pt, eta, dR, mass, sqrt_s
-#     (mass is for pairs of particles from the same group, sqrt_s is for all outgoing particles)
+# possible observables: pt, eta, dR, mass, m_inv, sqrt_s
+#     (mass is the mass of one particle, or of the summed momenta with "-sum-";
+#      m_inv is the invariant mass of a pair, applied to every pair the group
+#      can form; sqrt_s is for all outgoing particles)
+# An m_inv minimum is also used to bound the integration: where the pair is
+# what a propagator decays into, the cut becomes that propagator's minimum
+# invariant mass, so the forbidden region is never generated rather than
+# generated and thrown away.
 # for all cuts, min or max can be specified
 $cuts
 
--- a/madspace/include/madspace/phasespace/observable.hpp
+++ b/madspace/include/madspace/phasespace/observable.hpp
@@ -33,7 +33,7 @@ public:
         obs_delta_eta,
         obs_delta_phi,
         obs_delta_r,
-        obs_m_inv,
+        obs_pair_mass,
         obs_sqrt_s
     };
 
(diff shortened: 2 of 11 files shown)


theoheimel:
> There was no way to cut on every pair a group can form, and no way to cut at all on a group larger than two.

This is not true, the system is quite flexible. For example, `lepton-lepton-sum-mass` should apply an invariant mass cut for pairs of leptons and `lepton-jet-top-sum-mass` for triples consisting of a lepton, jet and top. Of course this is quite hard to find and currently not very well documented (even the AI didn't understand it), so this will also not be clear to the user.

Therefore I think it indeed makes sense to define a separate observable for pairs since this is a really common cut. We should change the naming though. I'd propose either going with `m_inv` and `pair_m_inv`, or `mass` and `pair_mass`. `mass` vs `m_inv` is confusing since both are invariant masses.

oliviermattelaer:
Renamed to `pair_mass` in 117797995, keeping `mass` as it is.

`obs_m_inv` -> `obs_pair_mass`, `kernel_obs_m_inv` -> `kernel_obs_pair_mass`, the run card key becomes `top-pair_mass.min = 800.0`, and the test file is now `madspace/tests/test_pair_mass_cut.py`. Rebuilt from source and checked against the built module: `observable='pair_mass'` constructs and reaches the phase-space floor, `observable='m_inv'` is now a `TypeError`. Full madspace suite 1512 passed, 2 skipped (`test_flow` / `test_mlp` need torch, `test_double_t` fails on `main` too).

You are right about the existing syntax, and the PR description was wrong to say there was no way to do this. Naming a group per member already gives one combination at a time, so `lepton-lepton-sum-mass` was always the invariant mass of a lepton pair and `lepton-jet-top-sum-mass` that of a triple. The cuts section of the run card now says so, and presents `pair_mass` as the shorter spelling of the first rather than the only one:

```
# possible observables: pt, eta, dR, mass, pair_mass, sqrt_s
#     (mass is the mass of one particle. Put "-sum-" before it and the momenta
#      are added up first: "lepton-sum-mass" is the mass of all the leptons
#      together, while naming a group per member gives one combination at a
#      time, so "lepton-lepton-sum-mass" is the mass of a lepton pair and
#      "lepton-jet-top-sum-mass" that of a lepton-jet-top triple.
#      pair_mass is the invariant mass of a pair, taken over every pair the
#      named groups can form, so "lepton-pair_mass" is the same cut as
#      "lepton-lepton-sum-mass". sqrt_s is for all outgoing particles)
```

## One thing your comment turned up, which I have not changed

If `lepton-lepton-sum-mass` is the pre-existing spelling of the same cut, then it ought to get the same phase-space bound. It does not. Measuring the floor `Cuts::m_inv_min` reports for a `min=500` cut on the top pair:

| cut | floor |
|---|---|
| `top-pair_mass` | 500.0 |
| `top-sum-mass` | 500.0 |
| `top-top-sum-mass` | **0.0** |

`Cuts::m_inv_min` only recognises a summed selection of exactly two particles (`idx.size() == 1 && idx.at(0).size() == 2`), which is the `top-sum-mass` shape. The `top-top-sum-mass` shape is one index list per group, so it falls through and gets the event-level filter without the bound. That is a behaviour change rather than a rename, so it is not in this push. Happy to add it here if you would like the two spellings to stay interchangeable.

Also, the accessor is still called `Cuts::m_inv_min()`. It pre-dates this branch and is public Python API (`Cuts.m_inv_min()`, `ColorOrderedMapping(m_inv_min=...)`), so renaming it reaches well past this PR. Say the word if you want the vocabulary uniform.

oliviermattelaer:
Tracked the phase-space bound gap as #121 so this can merge as it is.

PR #111, opened 2026-09-08
Numerically stable light-cone component in the massless fermion spinors, in every backend

oliviermattelaer:
`sqp0p3 = sqrt(p0+p3)` is used as **both a spinor component and a divisor** in the massless fermion wavefunctions. For a backward-moving massless fermion `p0+p3` is a cancelling difference of two ~|p| numbers, so it loses precision exactly where it is also divided by. The routines special-case the *exactly* backward parton (`px==py==0, pz<0`), which is accidentally right in the lab, but any transverse component lifts a momentum off that branch and the error becomes O(1).

How bad: rotating an event by **1e-6 rad** — an exact symmetry of |M|^2 — changes the matrix element by **1.06e-05** in every backend. In the density path that fed the boosted evaluation, a 1e-14 rad rotation changed a trace by a factor 1.6e5.

The fix takes the small light-cone component from the identity `p+ p- = pT^2` instead:

```
sqp0p3 = sqrt(pT^2 / (|p0| - p3))     for p3 < 0 and pT != 0
```

a sum of positives where the old form cancels. `p3 >= 0` is untouched, and the exactly-backward `pT == 0` case keeps its old exact branch, so the change is confined to `p3 < 0, pT != 0`.

## Scope: 44 sites in 10 files, every backend

fortran (`aloha_functions{,_fd,_loop}.f`, incl. `mp_` quad), C++ (`aloha_aux_functions{.cc,.fd_cc}`), madmatrix (`helas{,_fd}.h`, **SIMD and scalar**), python (`wavefunctions.py`), and the legacy `HELAS/{ixxxxx,oxxxxx}.F`. Routines: ixxxxx, oxxxxx, ixxxso, oxxxso, mp_ixxxxx, mp_oxxxxx, irxxxx, orxxxx, ixzxxx, oxzxxx.

Two sites deliberately left alone: `ilxxxx`/`olxxxx` take **complex** momenta, where `p(1)^2+p(2)^2` is not `|pT|^2` and the substitution is not an identity.

`ixzxxx`/`oxzxxx` had neither the `fpmax` guard nor a `pT==0` fallback — stock they return **inf/NaN** at small rotation angles, not merely inaccurate results. `irxxxx` had no `pT==0 .and. pz<0` branch at all; spin-3/2 now matches spin-1/2.

## Validation

64 phase-space points of `u u~ > d d~ g` at 1 TeV, exactly light-like, against an unpatched twin checkout.

**No-op on ordinary kinematics** (`|new/old - 1|` max): fortran 1.8e-15, fortran-FD 2.0e-15, madmatrix SIMD 1.9e-15, madmatrix scalar 2.1e-15, madmatrix-FD 2.1e-15, python 2.1e-15.

**Null-rotation regression**, `|M2(rot)/M2 - 1|` median at 1e-6 rad:

| | old | new |
|---|---|---|
| fortran | 1.06e-05 | 6.7e-16 |
| madmatrix | 1.06e-05 | 6.7e-16 |
| python | 1.06e-05 | 6.1e-16 |

**Cross-backend anchor** unchanged: madmatrix vs fortran on the same point, 1.49e-14 before and after.

For routines with no reachable matrix element, the exact spinor identity `|Sum oxxxxx(p2).ixxxxx(p1)|^2 = 2 p1.p2` with the RHS in exact rational arithmetic: `aloha_aux_functions.cc`, `.fd_cc`, `HELAS/ixxxxx.F` and `aloha_functions_loop.f` (double **and** `mp_` quad) all go 1.06e-05 -> 2.2e-16 at 1e-6 rad.

**Throughput:** a naive SIMD transcription cost -5.6% (a vector ternary evaluates both arms, so two `fpsqrt`). Restructured to select p+ before the square root: -0.7%, inside run-to-run scatter, bit-identical MEs.

## Also in this branch

`lhe_parser.project_massless_initial_state` — aMC@NLO writes LHE partons on the Monte-Carlo mass shell (`put_on_MC_mshell`), so an incoming light quark reaches the matrix element with `p^2 = (0.33 GeV)^2` while the ME treats it as massless. The inverse is closed-form (the map happens in the incoming pair's rest frame, so shat and the total initial four-momentum are exactly invariant) and the projection is exact: residual `p^2` = 0, `|shat'/shat - 1| <= 1.8e-13`, energy imbalance 0.

Wired explicitly into MadSpin (`get_density`, `get_density_batch`, `calculate_matrix_element`) and both reweight interfaces — deliberately **not** into `Event.get_momenta`, which has other callers. Scoped by the model's mass, not a PDG list. 5 new unit tests.

Boost invariance of `Tr rho` over 8000 events: median 3.8e-06 with 6.1% of events above 1e-3, to median 2.2e-15 with **0.000%** above 1e-3. Inclusive polarisation fractions are unchanged (f0 0.181632 -> 0.181633) — this was never a bias on an inclusive measurement; it moves the per-event density, up to 5.6% on `Tr rho`.

## Notes for review

- **IOTest goldens: nothing regenerated.** The only golden this branch makes drift is `FD_madmatrix/src/HelAmps_sm.h` (6 lines removed, 52 added, 10 hunks confined to the six patched spinor blocks) — the only golden in the tree containing `sqp0p3`. Both trees report 48 IOTests with the same pre-existing diffs. Regeneration is the maintainer's call.
- **`mp_ixxxxx` (MadLoop's quad-precision rescue) does not help** — the cancellation is already in the double-precision *input*, not in the arithmetic. An argument for fixing it in the routine rather than by promotion.
- **madmatrix's shipped `FPTYPE=m`** puts a ~3e-8 floor under everything and hides the improvement; all numbers above are `FPTYPE=d`.
- **Not built here:** CUDA/HIP (no GPU; the CUDA path is the `#else` of the same `#ifdef` and was measured as `BACKEND=scalar`), MadLoop end-to-end (COLLIER will not install), `mg7_v5` and `matchbox_cpp` (both broken on `main` independently of this branch).
- The final-state momentum projection is deliberately **not** included; the case for extending it, with numbers, is written up in `lc_fix/NOTES.md` section 14.


oliviermattelaer:
For polarization@NLO (and in link to madspin), looks like I need to be more precise and claude push me to move on this particular change independently of the @stloufra work on precision.

So this is the motivation to push forward such change independently.
But @stloufra if this is include in your "v" precision, we can also wait for this PR

stloufra:
It is included in the `v` but only for the `helas.h` and `helas_fd.h` without the case where 
```math
E - p_z < 0
```
for that case your PR sets E - p_z = 1. I am not sure if this is the correct way how to do that. Could you explain why this should be true?

oliviermattelaer:
This is something that Claude did only for the SIMD case:
```
volatile fptype_sv p0mp3DENOM = fpternary( p0mp3 > 0, p0mp3, 1. ); // hack: dummy p0mp3DENOM[ieppV]=1 if p0mp3[ieppV]<=0
      const fptype_sv p0p3stable = fpternary( ( pvec3 < 0. and p0mp3 > 0. ),
                                              ( pvec1 * pvec1 + pvec2 * pvec2 ) / (const fptype_v)p0mp3DENOM,
                                              (const fptype_v)p0p3 );
```

The justification was that the ternary operator evaluates both condition for each phase-space point.
So one should avoid that p0mp3DENOM = 0 even for events that do not use the denom branch.

In other word, you can set that value to any non zero value and it would be fine. This is just a trick to avoid a bad division

stloufra (approved):
Looks all good to me. Maybe we should steer away from the `// hack: dummy` comments in the future.

stloufra:
If CI passes we should be able to go through with the PR. In the merge I have kept the new approach with setting p0pm3DENOM to 1 as I think the readability is improved (the ternary in multiple steps not nested) even though the solution is same as before.

PR #114, opened 2026-09-09
MadSpin: absorb flavour-grouped leg re-orderings in the density and ME calls

oliviermattelaer:
## The symptom

`p p > z z j [QCD]` + MadSpin aborts, in **any** density spinmode:

```
MadSpinDegenerateWeight: the production spin-density matrix of process
'2 3 > 2 3 23 23' is identically zero -- Tr(rho_prod) = 0.0
```

It hits the `q q' > Z Z q q'` real-emission channel on ~0.5% of events — but one such event kills the whole run, after minutes of apparently healthy progress. This is **not** polarisation-specific: it reproduces on the plain unpolarised process.

## The cause

The generated flavour-grouped `matrix.f` deliberately exposes **one** leg ordering. For `P2_QQ_z0z0QQ`, `FLAV_TABLE` holds only the unordered flavour pairs — `(1,1) (1,2) (1,3) (1,4) (2,2) (2,3) (2,4) (3,3) (3,4) (4,4)` — and `M0_GET_FLAVOR_INDEX` returns a `0` sentinel for anything else, whereupon `GET_DENSITY` short-circuits to an identically zero density and `SMATRIX` returns 0. The fortran is behaving as designed; the python wrapper is meant to absorb the re-ordering.

It does not. And **neither `get_density` nor `calculate_matrix_element` absorbs it**, which was the surprise: the only source of alternative orderings is `get_all_momenta`, whose docstring says it returns *"all valid permutation of the final state"* — final-state slots only, and only among same-pdg particles with **different mothers**. In an aMC@NLO production event every final particle has mother `(1,2)`, so it returns exactly one identity permutation. `calculate_matrix_element`'s `for p in all_p` loop has a single element in practice, and `get_density`'s `assert len(all_p) == 1` was never in danger.

`get_all_momenta` also cannot repair this class even in principle: of the two failing orderings, one needs an **initial-state** swap, which it structurally never produces. (Stock's own diagnostic already said the scalar path was equally affected — *"the full production matrix element vanishes as well"*. Fixing only `get_density` moved the abort to `_combine_maxwgt`, "no usable bound", because the same zero is the accept/reject denominator.)

## The fix

A second permutation source: **relabelling legs that share a merged pdg inside the same block** (both incoming, or both outgoing), moving momentum *and* raw pdg together. That is a pure renaming of external lines, so every entry of the density is unchanged — verified numerically against the compiled ME (`scripts/probe_invariance.py`): on the events where all four relabellings resolve, the densities agree to **2.3e-12** relative, initial-state swap included.

Applied to `get_density`, `calculate_matrix_element` and `get_density_batch`, sharing one cache and one policy.

Design points:

- **The fast path is untouched.** Primary ordering, one fortran call. `_merged_flavor_groups` returns `[]` — disarming everything — unless a merged pdg repeats inside a block. The search runs only when the primary comes back as the sentinel, and a known repair is a dict lookup.
- **The resolution test reads the raw fortran INTER array, not `trace()`.** With a `hel_restriction` in force a *resolved* density can legitimately have a zero restricted trace, and `trace()` would mistake that for the sentinel.
- **`assert len(all_p) == 1` becomes a policy, not a deletion.** Multiple resolving `get_all_momenta` permutations are the genuine ambiguity and now go through `identical_particle_in_prod_and_decay` (`crash` raises the same message `calculate_matrix_element` raises; `first`/`average` as expected; `max` compares |unrestricted trace|, the density analogue of the scalar path's `abs(out) < abs(new_value)`). Merged-flavour relabellings are an exact symmetry, so the first that resolves wins silently.
- `need_raw_pdg` / `get_pdg` is re-derived **per permutation**, off the lab momenta, since the relabelling moves momentum and pdg together.
- Nothing resolves → the original zero array is returned and the caller's error is unchanged.

`get_density_batch` had the same inert `except` branch and the same assert, and never checked whether the block it got back was the sentinel. It is reached only for decay pools, where typical `z > l+ l-` MEs are disarmed — but a 1→3 decay with two same-sign merged quarks would arm it, so it is wired to the shared memo rather than left out.

## Validation

| | stock | with this PR |
|---|---|---|
| `p p > z z j [QCD]`, 5000 events + MadSpin | **aborts** (`Tr(rho_prod) = 0.0`) | **5000/5000 written, rc=0** |

Decayed σ = 4.7099e-02 pb against 1.041e+01 pb undecayed — ratio 4.524e-3 vs BR(Z→ee/µµ)² = 4.532e-3. No NaNs.

**No regression, byte-identical:** `p p > z z [QCD]`, 5000 events, RNG pinned — the decayed LHE is byte-identical (7 026 117 bytes, `cmp` clean) between stock and this branch, same 41 531 trials, same 0.1204 efficiency. That process is fully disarmed, which is exactly the point.

**Cost.** Over 5000 `p p > z z j` events: 99.28% disarmed, 0.32% armed-and-primary-resolves, **0.40% needing the repair**, across 6 distinct failing orderings. The run's own counters agree exactly — the search runs **6 times in the whole run** (~24 extra ME evaluations against 36 105 accept/reject trials); everything else is a dict hit. Per call: **+0.029 µs disarmed (0.2%)**, +0.088 µs armed. End to end the fixed build is never slower than stock.

597 MadSpin unit tests OK, including 18 new. Two files touched, no `template_files`, so **no IOTest golden drift**.

## Known limits

- With `nb_core > 1` the memo is per worker, so the search runs ≤6 times *per worker*.
- Only the `q q > z z q q` topology was exercised. Three or more same-merged-pdg legs in one block grows the relabelling count factorially; `_FLAVOR_RELABEL_MAX = 64` caps it, but that cap is untested against a real process.
- `ReweightInterface.calculate_matrix_element` carries the analogous merged-pdg logic and plausibly has the same hole. Not checked.
- Unrelated, found in passing: MadSpin's `set seed` does not reach `do_launch` in the `decay_events` path — `post_set_seed` seeds `random`, but `options['seed']` is falsy there and the banner records a different seed.


oliviermattelaer:
This should be move to a more internal standalone handling ...

PR #117, opened 2026-09-09
Merge upstream mg5amcnlo 3.8.0 into main

oliviermattelaer:
Merges upstream `mg5amcnlo/3.8.0` (`901080da4`) into `main`.

`mg5/3.8.0` brings MG5aMC 3.7.3 + the start of 3.8.0. Nine files conflicted; the rest merged cleanly.

## The main theme

mg7 (PR #115) and upstream 3.8.0 **independently implemented the same feature**: content-validation of the HEPTools CI cache, to stop a green `warm cache` run from publishing a cache with a tool silently missing. Upstream factored it into a reusable composite action (`.github/actions/check_heptools` + `check_heptools.sh`); mg7 did it inline in bash.

Upstream's version wins here: its checks are a **strict content superset** of ours (everything we checked — CutTools, IREGI, collier, ninja, the pythia8 header/binary — plus boost, YODA, hepmc3, the PDF sets, and `lib`/`lib64` variants), and it is one shared file instead of a dozen inline `test -f` blocks. Upstream also splits boost into its own `boost_cache` job, so a flaky boost download no longer discards the eMELA build.

mg7's orthogonal CI policy is re-applied on top:

- `ubuntu-24.04` only (no 22.04 matrix leg)
- `--ref "${{ github.ref }}"` on every `gh cache delete`
- the `github.ref_name == 'main' || 'test_ci' || workflow_dispatch` job gating
- `./bin/madgraph` and `input/mg7_configuration.txt` (`bin/mg5_aMC` does not exist in mg7 — upstream's new `boost_cache` and part of `emela_cache` would have hard-failed)
- the `looptools-<os>` entry in the `reset_heptools` delete list, which **upstream still lacks**
- `boost-<os>` added to the delete/cleanup lists, since boost now has its own cache

## Per-file resolutions

| File | Side | Why |
|---|---|---|
| `VERSION` | theirs | ours was stale at 3.7.2 |
| `UpdateNotes.txt` | theirs | verified strict superset — contains both mg7 BUG FIX entries verbatim, plus 4 more and the new 3.8.0 section |
| `Template/NLO/SubProcesses/genps_fks.f` | theirs | both sides made the identical `virtgranny_red` EXTERNAL fix, only continuation-line order differed |
| `tests/acceptance_tests/test_cmd_madevent.py` | theirs | both agree `cross=15.73`; the `error=0.04` set on 26 Jun was loosened to `0.514` upstream two days later in `75d821554 "fix CI test"` |
| `tests/acceptance_tests/test_cmd.py` | theirs | `--no_open` |
| `madgraph/interface/madgraph_interface.py` | theirs **+ ours** | upstream's `do_draw` rework (`--no_open`, `--merge`→PDF) had already won everywhere else in the file via auto-merge; mg7's `--generate_only` is kept as an alias so the acceptance tests and the flag both keep working |
| `.github/actions/restore_heptools/action.yml` | theirs | see above |
| `.github/workflows/acceptancetest.yml` | **ours** (both hunks) | net diff vs `main` is zero — see below |
| `.github/workflows/warm_cache.yml` | mixed | 18 hunks, resolved individually |

### Why `acceptancetest.yml` takes ours

Two traps in that file:

1. The `acceptancetest_uncovered_merged:` job *header* is unconflicted, but upstream's side of the conflict is the body of `acceptancetest_heft` (a job mg7 deliberately pruned, and whose header the auto-merge had already deleted). Taking upstream would have grafted the heft steps under our job's name **and deleted five mg7 jobs** — `uncovered_merged`, `uncovered_madloop`, `improve_ps_at_rest`, `check_madevent_me_vs_standalone`, `check_mlm_rwgt`.
2. Upstream appends `acceptancetest_delphes_parallel:` at EOF, but that job already exists intact at line 1288 — a **duplicate YAML key**, which kills the whole workflow.

## Two problems found outside the conflicts

Both are clean auto-merges that produce broken YAML, so they would not have shown up as conflicts:

- **`looptools_cache` had the check twice.** Git merged our inline `for f in …` loop (5-space indent) and upstream's `check_heptools` step (7-space indent) into the same job, same step name. `warm_cache.yml` **did not parse**. Resolved to upstream's action, with its `if: cache-hit != 'true'` dropped so it still runs on the cache-hit path — that was the deliberate behaviour in the mg7 comment (an already-stored incomplete cache gets reported rather than silently reused).
- **`heptools_cache` had a duplicate `if:` key.** Upstream's block carries its own `if: always() && …` while mg7 already has one on that job. Removed upstream's; the `always()` semantics live in ours.

Also: `Restore lhapdf (fallback when pythia8 and emela failed)` was dropped by the auto-merge. That is fine — upstream restores lhapdf unconditionally first, which subsumes it.

## Verification

- no conflict markers in any file touched by the merge
- all workflow/action YAML parses: 13 jobs in `warm_cache.yml`, 73 in `acceptancetest.yml`
- no duplicate step names or undeclared `needs.<job>` references introduced (the one duplicate in `acceptancetest_density` predates this merge on `main`)
- every `uses: ./.github/actions/X` resolves to an existing `action.yml`
- no `bin/mg5_aMC` / `mg5_configuration.txt` references left anywhere under `.github/`
- Python files parse; `check_heptools.sh` passes `bash -n`

The acceptance tests themselves were **not** run locally — they need a full CI environment, so CI on this PR is the real check.


oliviermattelaer:
## CI triage on the first run

### Fixed in 9959eefbf

`unittest_10` and `unittest_madspin_sequential` — two stale test doubles. Both tests arrive **with** `mg5/3.8.0` and are written against upstream's `interface_madspin.py`, but the merge kept mg7's richer implementations:

- `TestReopenDecayPoolFallback._Stub` had no `_count_pool_events`. mg7's `_reopen_decay_pool` calls it on the own-file fast path to size the owner's ~10% trim; upstream's version doesn't need it. → `AttributeError: '_Stub' object has no attribute '_count_pool_events'`
- `TestMaxWeightScanForksEveryWorkerAnOwnerCanName`'s `Worker._open_refill_slice(self, decay_dir, gen, owner)`. mg7's `_worker_refill` passes the channel cross-section as a fourth argument. → `takes 4 positional arguments but 5 were given`

Only the test doubles moved; no production code touched. Both classes pass locally (8 tests, OK).

### Not ours

`acceptancetest_delphes_parallel` — `apt-get update` failed on Google Chrome's apt repo inside `restore_delphes`:

```
E: Failed to fetch https://dl.google.com/linux/chrome-stable/deb/dists/stable/main/binary-amd64/Packages.gz  Hash Sum mismatch
```

A runner-side flake, re-run with the push.

`IOtest_TIR` / `IOtest_matchbox` — deliberately not touched. `running IOTests` is also red on `main` at the moment, so at least part of that drift predates this merge. Golden regeneration is a maintainer call.

### Open — needs a physics decision, deliberately left failing

`madspin_factory (test_short_madspin_zz)`:

```
onshell_decay_chain = 0.164   onshell_density = 0.2159   rel = 0.240 > close_rel_tol = 0.15
```

This is **not** a merge artifact. It is upstream commit `69bc475b6` ("MadSpin: extract `_decay_symmetry_factor` and pin it in the unit tests"), which the merge took correctly. It replaces mg7's inline symmetry factor — keyed on the drawn *final state* — with one keyed on the *channel*, applied only when the channels were **dealt** (`ms_positional`).

`test_short_madspin_zz` uses `decays=['z > l+ l-']` with `l+ = e+ mu+`: one merged pool, both Z bosons drawing from it independently, so `ms_positional` is False and the new rule applies **no factor**. That is exactly the case the new docstring names as the old bug — mg7 halved the denominator for mixed-flavour trials, giving `e+e-mu+mu- : 4e : 4mu` of 4:1:1 in place of 2:1:1.

So `onshell_density` moved on purpose and `onshell_decay_chain` (unchanged, 0.164) is now the mode out of step. What is caught in the middle is rule 2a of `assert_efficiency_ordering` — "the two on-shell routes agree within `EFF_TOL`".

Worth noting that `assert_multiplicities_consistent` passed and **cannot** discriminate here: the mean e+ count is 1.0 under both 4:1:1 and 2:1:1. The discriminating observable is the four-lepton flavour composition, which nothing in this test currently checks.

Three ways out, none of which should be picked by CI triage:

1. the legacy `onshell_decay_chain` route carries the same double-count and needs the equivalent fix, after which the two agree again at ~0.216;
2. the new channel-keyed rule is wrong for this configuration;
3. rule 2a genuinely does not apply to a merged-decay-line process and needs a documented carve-out.

Leaving this red pending that call.

oliviermattelaer:
## CI green after the re-merge — 278 pass, 0 fail

Every failure from the first round is resolved.

| First round | Now | How |
|---|---|---|
| `madspin_factory (test_short_madspin_zz)` | pass | upstream `b12a8a45d` — the legacy `onshell_v1` decay-chain route carried the extra identical-particle factor, not the density |
| `unittest_10`, `unittest_madspin_sequential` | pass | the two stub fixes, superseded by upstream and merged cleanly |
| `acceptancetest_delphes_parallel` | pass | the apt `Hash Sum mismatch` flake, gone on re-run |
| `IOtest_TIR`, `IOtest_matchbox` | pass | **not touched** — no golden regenerated. They were almost certainly the cache-window bug `3513ae7bc` describes: `running IOTests` was red on `main` at the same time, and upstream saw the same signature (13 acceptance jobs red, then green on a re-run of the identical commit) |
| `acceptancetest_96` | pass on re-run | pre-existing concurrency race, see below |

### acceptancetest_96 — a pre-existing race, not this merge

`test_madspin_ON_and_onshell_atNLO` failed once and passed on a plain re-run of the same commit. The mechanism is worth recording because the symptom points nowhere near the cause.

The decay-ME build died with two errors, both at **column 72**:

```
aloha_functions.f:1143:72  ...1d-99))*rHalf   -> Error: Missing exponent in real number
cuts.inc:6:72              ...mue_over_ref... -> Error: Symbol 'mu' has no IMPLICIT type
```

Fixed-form Fortran truncating at 72 columns, i.e. `madspin_me/Source/make_opts` was not in effect. Three independent symptoms in the same compile agree:

- `f77` was used — make's built-in `$(FC)`, so `FC=$(DEFAULT_F_COMPILER)` never fired;
- `-ffixed-line-length-132` absent, which `make_opts` adds unconditionally for non-ifort;
- the library target was `../../lib/libdhelas.` with an **empty** `$(libext)`, and `libext` is only defined in `make_opts`.

The job log shows **five concurrent compile failures against one shared path**, `/tmp/amc5z2fnpa9/MGProcess/madspin_me/Source`. So: five processes build the same `madspin_me`; `make_opts` is populated with `shutil.copy` (`madgraph/__init__.py:49`, `madgraph_interface.py:3333`), which truncates the destination before writing; and `replace_make_opt_f_compiler` (`export_v4.py:3323`) swallows the resulting `IOError` behind a `logger.info`. A process compiling inside that window gets pure make defaults.

An empty-but-present `make_opts` fits better than a missing one: make parsed it without complaint, and `FC=$(DEFAULT_F_COMPILER)` lives in the body that was truncated away, so `FC` stayed `f77`.

Nothing in this merge touches that path — `MadSpin/interface_madspin.py` has two hunks, both pure weight arithmetic, and `amcatnlo_run_interface.py`'s fastjet guard writes `me_dir/Source/make_opts`, not `madspin_me/Source`, with a different failure mode. Left unfixed here deliberately: it is a real bug but out of scope for a merge, and it deserves its own change.

PR #123, opened 2026-09-10
Tutorial mode: a step engine, a menu, and ten new tutorials

oliviermattelaer:
Rebuilds tutorial mode as a real mode, and adds ten tutorials to the two that existed.

## Why

The old tutorial mode was a logger trick rather than a mode: three text modules, three loggers, and a command-name → text lookup run after every command. Because the lookup key *was* the command name, a tutorial could teach `generate` exactly once — which rules out every tutorial worth adding, since they are all sequences of several `generate` or `import model` steps.

## The engine

`madgraph/interface/tutorials/` holds a step engine. A tutorial is an ordered list of steps; a step is triggered by a command and prints the lesson that follows it. Two ordering policies: `free` (any step, any time — what the old tutorials effectively had, kept for the ported ones) and `sequence`, which lets a tutorial teach the same command repeatedly with a different lesson each time.

While a tutorial runs, `TutorialMixin` is spliced in front of the **instance's** class. That is deliberate and it is the only thing that works: cmd dispatch resolves `do_XXX`, `postcmd` and `default` on the instance, while `Switcher.self.cmd` is consulted only for the methods Switcher forwards — so wrapping `self.cmd` would reach neither `postcmd` nor any command the mixin adds. Splicing at the instance also means a tutorial survives the automatic LO ↔ NLO interface switch with no hook in `change_principal_cmd` at all, which is verified: `syntax` walks straight through `generate p p > t t~ [QCD]` into the aMC@NLO interface and carries on.

Because the mixin is only in the MRO while a tutorial runs, tutorial mode now costs nothing when it is off — better than before, which ran three `getattr`/`try`/`except` pairs after every command whether or not anyone was learning.

## Commands

`tutorial` with no argument opens a menu, grouped Basic / Advanced / Exercises. `tutorial list`, `status` and `help` are new, as are `hint`, `solution`, `next`, `repeat`, `back` and `skip` while one runs (also accepted as `tutorial hint` and so on). None of them execute anything — the user always types the command.

Primary names are `lo` / `nlo` / `madloop`, with `MadGraph5` / `aMCatNLO` / `MadLoop` kept as hidden aliases, so existing scripts and proc cards are unaffected.

`crash_on_error` is suspended while a tutorial runs and restored on `tutorial stop`. A tutorial is a place to make mistakes; having a typo tear down the session is the opposite of what the mode is for.

## The tutorials

**Basic** — `lo` (rewritten: from a cold start to an event sample on madspace, which is what `output` produces by default), `nlo`.

**Advanced** — `syntax`, `mg7`, `madevent`, `model`, `bsm`, `standalone`, `decays`, `madloop`, `checks`.

**Exercises** — `exercises`: MG7 asks, you answer, and it names the mistake you made. Checks inspect the state the command produced rather than the text typed, so `p p > t t~ QED=0` and `p p > t t~ QCD=2 QED=0` both pass. Wrong answers get diagnosed — `QED=2 QCD=2` when asked for the interference is told those are amplitude orders; `QCD^2=2` is told a bare `=` means `<=`; a `>` where a `,` was meant is told the separator is a comma.

The Advanced and Exercises headings say the content is AI-generated and not yet validated. `nlo` and `madloop` are the pre-2026 hand-written text, ported near verbatim.

## Two changes outside the tutorials

- **The default `parton_shower` is now PYTHIA8**, not HERWIG6 (`banner.py`). HERWIG6 does not build with a current gfortran, so the shipped default was one nobody could run. Two acceptance tests pinned the old default and are updated. Note `--std=legacy` is *already* applied to that build and is not sufficient — `SPLIT` became a Fortran 2023 intrinsic, so `jimmy.f`'s calls are checked against the wrong signature. That is filed separately; patching vendored Herwig source is a maintainer call.
- **`aloha/create_aloha.py` dropped a dead import** of the tutorial text — provably unused (checked with an AST walk), and it made aloha import `madgraph.interface`.

`misc.TMP_variable` gained dict support so it can reach `options['crash_on_error']`; the attribute path is untouched.

## Method

Every command a tutorial asks for was run before it was written. That caught, among others: MG5 does **not** add `QED=0` when you give no orders (it searches for the lowest `WEIGHTED = QCD + 2*QED`, and the old text had been claiming otherwise for years); `[madnis] enable` is `"auto"`, not off, and auto also sizes the networks; there is no `train_madnis` driver, training is a phase inside `bin/generate_events`; `run_card.toml` lives in `Cards/`; `check` does not load a model for you, unlike `generate`; and a step must never ask for a command that blocks on an editor.

## Testing

107 tutorial tests. The ones that carry weight are the self-checking ones: every sequenced tutorial is walked by typing its own solutions, every exercise must pass its own solution, and every named mistake must be reachable — feed the wrong answer, assert that diagnosis fires. A mistake that stops firing means the syntax moved.

Full unit suite compared against pristine `main`: 1648 vs 1549 tests (+99), with **identical** failure and error sets — 2 failures and 94 errors on both sides, all pre-existing.

## Not done

`run`, `performance` and `install` tutorials; child-interface propagation (so a tutorial could teach inside an interactive `MadEventCmd`); retiring `tutorial_text*.py`. `docs/tutorial-mode-plan.md` records the design, what landed, and why `run` in particular is deliberately unwritten.


oliviermattelaer:
So I would propose to merge this with the unvalidated tutorial for the alpha and then spend some time trough those tutorial/excercice to make them better/correct/...

oliviermattelaer:
let's merge #89 first, such that I can update the parser of the systematics results to match the new output style

theoheimel:
I only checked the LO tutorial so far. One thing we should definitely change is the amount of text. I think every step should fill half the screen at most, otherwise it is a bit overwhelming (and people won't read it).

oliviermattelaer:
So I have merged here 
#89 in order to have the correct parsing of systematics
#131 in order to allow to have the same integration (usefull for tutorial but not only) as madevent 

Then I have reduce the amount of text (and present some of it at a better moment)
and improve the history command to have the full history (including the set command used to edit the cards)

PR #128, opened 2026-09-10
Fix UFO custom propagators in the C++/madmatrix backend

oliviermattelaer:
## What

`import model SMEFTsim_U35_MwScheme_UFO; generate p p > t t~ NP=1; output` died with

```
aloha/create_aloha.py:593  propagator = getattr(self.model.propagators, propa)
AttributeError: 'NoneType' object has no attribute 'propagators'
```

That turned out to be the first of four bugs on the same path. A UFO model can attach a **custom propagator** to a particle — SMEFTsim uses them for the finite-width corrections of the t, W, Z and H — and none of these paths had ever been exercised, because they all sit behind `explicit_combine=True`, i.e. the C++/madmatrix backend only. `output madevent` and `output standalone_fortran` were never affected: fortran picks the parameters up from `coupl.inc`.

## The four bugs

**1. `aloha/create_aloha.py` — the builder never got the model.**
`AbstractALOHAModel.compute_subset` built the explicit-combine builder as `CombineRoutineBuilder(l_lorentz)`, dropping the optional `model` argument. Every other builder in the file passes `self.model`; this one call site was missed back in `e8c2a673b` (2012). It only bites when a `P<name>` custom-propagator tag needs `model.propagators`, and the `assert model` in `AbstractRoutineBuilder.__init__` is commented out, so nothing caught it in between.

**2/3. `madmatrix/model_handling.py` — two missing declaration branches.**
`MadMatrixALOHAWriter.get_declaration_txt` had no case for declaration type `fct` (`KeyError: 'fct_v'`) or `parameter` (`KeyError: 'parameter_v'`). `aloha_writers.ALOHAWriterForCPP` handles both; the GPU override never copied them.

A model parameter used inside a routine body is now defined from the `Parameters` class — `constexpr` under `MGONGPU_HARDCODE_PARAM`, the host singleton otherwise, which is why HRDCOD=1 is required on GPUs (there is no `-rdc=true` in the build, so a cross-TU device symbol is not an option either).

`FCTn` was assigned but never declared, and the single-`%s` format could not express `pow`. It is now emitted as `const cxtype_sv FCTn = ...` — which is exactly the form `write_combined_parts_cc` already greps for when merging the structures of an assembled routine — and `pow` maps to a new `cxpow` helper.

**4. `mgOnGpuVectors.h` — missing mixed operators.**
The propagator expressions produce `cxsmpl<double> + cxtype_v`, which had no overload: `*` and `/` already had both orders, `+`/`-` did not. Added `cxtype ± cxtype_v` (both orders) and `cxtype_v / cxtype`, plus `cxpow` (no `std::pow` for `cxtype_v`; the exponent always comes from the UFO propagator and is a small positive integer).

`cxpow` deliberately lives in `mgOnGpuVectors.h` rather than `cpp_hel_amps_h.inc` — the latter is an `IOTestFDGauge` golden, and this way no golden moves.

## Validation

- **Cross-backend, custom propagator**: `u d~ > e+ ve NP=1` (SMEFTsim W1 propagator, `dWW`). C++ `check_sa.exe matrix 1000`, simd_128 / `FPTYPE=d`: `7.1300507507539050e-03`; fortran `./check`: `7.1300507507539058e-03`. **1 ulp.** (The default `FPTYPE=m` build differs at 1e-8, as expected for mixed precision.) Builds clean under `-Wall -Wshadow -Wextra`.
- **No regression**: SM `g g > t t~` regenerates a **byte-identical** `HelAmps_sm.h`, and `check_sa.exe matrix 1000` gives `5.9262614118518897e-01` both before and after — so the new operators introduce no overload ambiguity anywhere.
- `tests/parallel_tests/test_aloha.py` (124 tests) and `IOTestFDGauge` pass.
- Full `./tests/test_manager.py -p U` (1551 tests): failure set **identical** to a `HEAD~1` baseline run on the same machine (2 failures / 94 errors, all pre-existing — madspin density, numpy pools, `test_output_standalone_cpp_refuses_loop_induced` which expects the removed `standalone_cpp`, ...). Note the suite shares a gitignored `UNITTEST_proc/` between runs, so the first run in a fresh tree reports a few extra order-dependent failures either way.

## Known limitation (unchanged, now explicit)

`p p > t t~ NP=1` still cannot be generated for the C++ backend. The Higgs propagator needs `mdl_dWH`, which depends on alphaS (through `gHgg1`/`gHgg2`) and is therefore recomputed event by event inside `computeDependentCouplings_fromG` — there is no addressable value a HelAmps routine could read, in any build mode. Instead of emitting code that cannot compile, the writer now raises:

> The custom propagator used by routine FFS2PH1_3 needs the model parameter mdl_dWH, which depends on alphaS and is therefore recomputed event by event. The C++/CUDA backend can only read alphaS-independent parameters inside a HelAmps routine, so this process can only be generated with "output madevent" or "output standalone_fortran".

Supporting it means threading the value in as an extra routine argument through the dependent-coupling buffer (new `DependentCouplings_sv` member + signature change in the ALOHA writer + matching change at every call site in the ME writer). That is a feature, not a fix, so it is out of scope here.

Separately, and unrelated to this PR: `HRDCOD=1` does not build for SMEFTsim at all, because `Parameters.h` initialises `constexpr double mdl_propCorr` with `std::abs(...)`, which is not a constant expression (6 errors, all in `Parameters.h`). Worth its own issue if the hardcoded-parameter path matters for this model.

### DIFF
--- a/aloha/create_aloha.py
+++ b/aloha/create_aloha.py
@@ -1233,7 +1233,7 @@ class AbstractALOHAModel(dict):
                         self[(lorentzname, outgoing)].add_combine(list_l_name[1:])
                     continue
 
-                builder = CombineRoutineBuilder(l_lorentz)
+                builder = CombineRoutineBuilder(l_lorentz, self.model)
                                
                 for conjg in request[list_l_name[0]]:
                     #ensure that routines are in rising order (for symetries)
--- a/madgraph/iolibs/template_files/madmatrix/mgOnGpuVectors.h
+++ b/madgraph/iolibs/template_files/madmatrix/mgOnGpuVectors.h
@@ -10,6 +10,7 @@
 #include "mgOnGpuCxtypes.h"
 #include "mgOnGpuFptypes.h"
 
+#include <cassert>
 #include <iostream>
 
 //==========================================================================
@@ -370,6 +371,18 @@ namespace mg5amcCpu
     return cxtype_v( a.real() + b, a.imag() );
   }
 
+  inline cxtype_v
+  operator+( const cxtype& a, const cxtype_v& b )
+  {
+    return cxtype_v( a.real() + b.real(), a.imag() + b.imag() );
+  }
+
+  inline cxtype_v
+  operator+( const cxtype_v& a, const cxtype& b )
+  {
+    return cxtype_v( a.real() + b.real(), a.imag() + b.imag() );
+  }
+
   inline const cxtype_v&
   operator+( const cxtype_v& a )
   {
@@ -416,6 +429,18 @@ namespace mg5amcCpu
   operator-( const fptype_v& a, const cxtype& b )
   {
     return cxtype_v( a - b.real(), fptype_v{} - b.imag() ); // IIII=0000-b.imag()
+  }
+
+  inline cxtype_v
+  operator-( const cxtype& a, const cxtype_v& b )
+  {
+    return cxtype_v( a.real() - b.real(), a.imag() - b.imag() );
+  }
+
+  inline cxtype_v
+  operator-( const cxtype_v& a, const cxtype& b )
+  {
+    return cxtype_v( a.real() - b.real(), a.imag() - b.imag() );
   }
 
   inline cxtype_v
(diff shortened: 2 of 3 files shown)


oliviermattelaer:
@Qubitol , so this fixes the proppagator not handle correctly for madmatrix.
But as said above, this is not enough to fix all the issues.

This PR fixes two others, but this particular model has a propagator that depends of one additional coupling (running in top of that) which we would need to think on how to handle ...

I guess that it is still worth the fix.

Qubitol (approved):
I see, for now let's merge this and then we'll see.
I'll create an issue with the problem still to solve, that I will take care of whenever I will have time to take care of the thread safety (which is going to make every parameter local, so maybe it will be easier to then consider the alphaS-dependent parameters).

PR #132, opened 2026-09-10
Fix the LHE intermediate-particle record in the mg7 output

oliviermattelaer:
Three defects in what madspace writes for LHE status-2 (intermediate) lines. Weights, momenta and cross sections are untouched throughout — but MadSpin, Rivet routines keyed on a boson, and any decay-chain analysis read these rows.

Found while checking the W sign for `p p > w+ 1-2j` and `p p > e+ ve +0-2j`. **W+jets was already correct** (the W is external there, status 1, and `subprocesses.json` has no W propagator at all); the problem is entirely on the `e+ ve` side.

## 1. Resonances built from the beam side got the anti-particle pdg

Madgraph records, on the leg a vertex creates, the pdg of the line flowing *into* the legs that were combined. That is already the decay orientation while those are all final state — but a line holding **every** initial leg is the one madspace roots the other way round, so what belongs in the LHE is the anti-particle.

```
p p > e+ ve       every 0-jet event wrote  -24  decaying to e+ ve
p p > e+ ve h     one channel carried the production W- and the decaying W+
p p > t b~        -24
```

`madevent` does not have this: it re-orients first via `get_s_and_t_channels`.

Fixed in `OneProcessExporterMG7.propagator_pdg`, now used by both propagator writers. Flipping on *any* initial leg instead would also hit t-channels, which never become decays — that wider rule wrongly flipped a top in a `t t~` decay chain.

## 2. Nested resonances lost their parent link

A resonance sitting under a plain internal line kept mother `(1,2)`, because `child_prop_mask` only ever named *direct* children — a `q q~ > Z > l l(-> l W)` diagram wrote the W as a child of the beams and left the Z with the wrong daughters. Two more ways the same link broke, found by reading the loop rather than from a failing event:

- `complete_event_data` mapped `prop_index` to row by counting set mask bits, which lands on the wrong row as soon as a propagator in between is resonant without being a child;
- the mask is decided once from `e_min < mass` while resonance is re-decided per event from the BW window, so it could name a child this event never wrote.

## 3. `bw_cutoff` reversed past `mass/width`

`m2 > m_min*m_min` with an unclamped `m_min = mass - bw_cutoff*width` turned a below-zero window into a large positive floor, so *raising* `bw_cutoff` wrote *fewer* resonances. `m_min` is now clamped at 0. The phase-space side never needed this — it already guards with `e_min > 0.` and `std::max`.

## Verification

Every comparison has the seed pinned, so only the intended columns may move.

| | change |
|---|---|
| (1) `p p > e+ ve +0,1,2j` | exactly 14150 lines, all `-24` → `24`, rest of line identical |
| (1) `p p > e+ ve mu- vm~` | byte-identical (219 371 lines) |
| (2) `p p > e+ ve mu- vm~`, 200k events | 27 lines of 2 198 022, each a W's mother `1 2` → `3 3` |
| (3) `bw_cutoff` 15 and 30 | byte-identical; row count monotonic again — 393848 / 398021 / 439769 / 443691 at 15 / 30 / 150 / 1000 |

Bad-mother counts on the `p p > e+ ve mu- vm~` reproducer went 7 → 0 at the default `bw_cutoff = 15` and 27 → 0 at 30. Both arms were rebuilt from source at identical flags for that comparison — an older install differs by a couple of borderline resonances purely from build flags, which looks like a behaviour change and is not.

The check used throughout is stronger than a charge test: for every status-2 line, its declared daughters' four-momenta must sum to its own. Clean across `p p > e+ ve +0,1,2j`, `p p > w+ j / j j`, `p p > e+ ve mu- vm~` and a `t t~` decay chain — and at `bw_cutoff = 1000`, where nearly every propagator is written and resonances nest everywhere.

`test_color_flow_consistency_mg7` and `test_add_time_of_flight_mg7` pass (the former zeroes all masses, so every propagator's colour is validated through the real C++ walk). `test_group_subprocess_mg7` skips on my machine for a missing LHAPDF data path, and `test_w_production_with_ms_decay_mg7` fails identically on and off this branch (no numpy in the interpreter MadSpin spawns) — both pre-existing and unrelated.

## Note for reviewers

This touches madspace C++, so `python3 madspace/install.py --source --yes` is needed to pick it up — the launcher only installs when `madspace/install/` is missing outright.

### DIFF
--- a/madgraph/iolibs/export_mg7.py
+++ b/madgraph/iolibs/export_mg7.py
@@ -226,6 +226,33 @@ class OneProcessExporterMG7(export_cpp.OneProcessExporterCPP):
             leg_sets.append(downstream)
         return leg_sets
 
+    def propagator_pdg(self, leg, leg_set):
+        """Signed pdg id of the internal line `leg`, oriented the way the
+        phase-space topology reads it: flowing away from the initial state.
+
+        Madgraph records, on the leg a vertex creates, the pdg of the line
+        flowing into the legs that were combined to make it -- `leg_set`.
+        That is already the decay orientation while those are all final
+        state, but a line holding *every* initial leg is the one madspace
+        roots the other way round: its decay products are the complementary
+        legs, so what belongs in the LHE is the anti-particle. Without this
+        the s-channel W+ of `p p > e+ ve` is written as a W- decaying to
+        e+ ve. A leg set holding only some of the initial legs is a
+        t-channel, which never becomes a decay and is left as madgraph put
+        it.
+
+        Only colour singlets actually depend on this: for a coloured line
+        lhe_output.cpp's compute_decay_color infers the orientation from the
+        colour flow and flips the pdg back itself.
+        """
+        part = self.model.get_particle(leg.get("id"))
+        if part.get("self_antipart"):
+            return part.get("pdg_code")
+        sign = 1 if part.get("is_part") else -1
+        if sum(1 for name in leg_set if name.startswith("i")) == self.n_initial:
+            sign = -sign
+        return sign * part.get("pdg_code")
+
     def diagram_propagator_pdgs(self, diagram, channel_leg_sets, sym_perm):
         """Signed pdg id of each internal line of `diagram`, reordered to
         match `channel_leg_sets` (the order used for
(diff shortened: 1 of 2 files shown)


theoheimel (approved):
We should probably add some more tests for this sort of thing in the future. For now, it looks good to merge.

PR #139, opened 2026-09-11
Compile madspace in parallel when ninja is missing

oliviermattelaer:
## The bug

`madspace/install.py` builds through `pip` → scikit-build-core, and **scikit-build-core runs `cmake --build` with no `-j` and never sets `CMAKE_BUILD_PARALLEL_LEVEL`** (only its legacy setuptools shim has a `--parallel`). So the job count has to come from the environment, and nothing supplied one:

- **with ninja**: invisible — ninja parallelises on its own;
- **without ninja**: scikit-build-core falls back to the `Unix Makefiles` generator, and `make` with no `-j` compiles one file at a time.

That is the "compiling madspace with GCC is not done in parallel" report.

## The fix

`install.py` gains `-j/--jobs N` (default: every available core) and sets `CMAKE_BUILD_PARALLEL_LEVEL` in the environment handed to pip. That variable is honoured by `cmake --build` for *either* generator and, being an environment variable, is inherited by the nested OpenBLAS `ExternalProject` too. An explicit value already in the caller's environment still wins.

From the MadGraph prompt the count is the `nb_core` option:

- `install madspace` appends `-j <nb_core>` unless the line carries its own `-j`/`--jobs` (also documented in `help install`);
- the in-process `launch mg7` passes it through `bootstrap.ensure_madspace(..., jobs=)`;
- `nb_core` is read straight from `self.options`: `set_configuration` already resolves an unset value to the machine's core count (via `set2_nb_core`);
- the standalone `bin/generate_events` bootstrap runs before any option is read, so it keeps the all-cores default.

## Measurements

18-core macOS, no ninja, generator confirmed `Unix Makefiles`, clean build dir each time:

| build | wall |
|---|---|
| serial (pre-fix behaviour, `-j 1`) | **2m50s** |
| new default (`-j 18`) | **1m05s** |

2.6x rather than more because there are only 58 objects and two dominate the critical path (`src/cpu/runtime.cpp`, `src/python/madspace.cpp`). Propagation through the full `pip --no-build-isolation` → scikit-build-core → `cmake --build` chain was checked separately on a 6-job probe project: 16.4s → 3.0s.

CI is untouched: `.github/actions/build_madspace` calls `pip` with build isolation, which pulls the ninja wheel.

## Tests

6 new tests in `tests/unit_tests/interface/test_mg7_launch.py` (nb_core forwarded by `install madspace` and by the mg7 launch path, an explicit `-j` left alone, `-j` reaching the installer command line). `./tests/test_manager.py -p U test_mg7_launch` → 46/46 OK.

Note for whoever runs the wider set: `test_manager.py -p U test_tutorials test_mg7_launch` reports 18 errors (`module 'madspace' has no attribute 'SOURCE_HASH'`) both with and without this branch — `tutorials/lo.py:153` does a bare `import madspace` that can bind a stale copy and caches it for `launch.py`. Pre-existing, tracked separately.

### DIFF
--- a/madgraph/interface/extended_cmd.py
+++ b/madgraph/interface/extended_cmd.py
@@ -1740,12 +1740,17 @@ class Cmd(CheckCmd, HelpCmd, CompleteCmd, BasicCmd):
         if self.log:
             logger.info("History written to " + output_file.name)
 
+    def get_nb_core(self):
+        """Number of cores available for parallel tasks: the nb_core option,
+        resolved to the machine's core count when it was left unset."""
+        import multiprocessing
+        if not self.options.get('nb_core') or self.options['nb_core'] == 'None':
+            self.options['nb_core'] = multiprocessing.cpu_count()
+        return int(self.options['nb_core'])
+
     def compile(self, *args, **opts):
         """ """
-        import multiprocessing
-        if not self.options['nb_core'] or self.options['nb_core'] == 'None':
-            self.options['nb_core'] = multiprocessing.cpu_count()
-        return misc.compile(nb_core=self.options['nb_core'], *args, **opts)
+        return misc.compile(nb_core=self.get_nb_core(), *args, **opts)
 
     def avoid_history_duplicate(self, line, no_break=[]):
         """remove all line in history (but the last) starting with line.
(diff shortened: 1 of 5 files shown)


oliviermattelaer:
@theoheimel I propose to not include this for the release, but doing it as soon as the alpha is out

theoheimel (approved):
Looks good, let's merge this.

PR #141, opened 2026-09-11
Fix gpu ddm

oliviermattelaer:
Fixing the DDM crash on GPU


oliviermattelaer:
@theoheimel I'm confused why it says two commit on this PR, it should have been one ... since the first commit, I thought that you included it in main already ...
 Let me try to clean that to avoid surprise

PR #148, opened 2026-09-12
Let a sqrt_s minimum bound the integration instead of only filtering it

oliviermattelaer:
## What this fixes

`madspace/src/phasespace/phasespace.cpp` computed the floor that a `sqrt_s` minimum puts on s-hat and then dropped it on the floor: `sqrt_s_hat_min` and `s_hat_min` were assigned and never read again. The only thing `[cuts] sqrt_s.min` ever did was multiply finished phase-space points by zero, so every point below the cut was generated and thrown away. The `pair_floor` block added above it in ec820297a — the bound that applies when no single propagator carries the cut pair — fed the same dead variable.

The fix hands the floor to the sampler the way a `pair_mass` cut already is, by raising the **root** propagator's `e_min`. `update_mass_min_max` carries `e_min` into the `s_min` the luminosity's `Invariant` is sampled from, and that `Invariant`'s Jacobian follows whatever range it is given — so the integral is unchanged and only the waste moves. Guarded to `_map_luminosity`: a leptonic collision has s-hat fixed at s_lab and `chili` reconstructs it from momenta it has already generated, so in neither case is there a range to narrow, and a floor at or above the beam energy would be an empty range. Those cases stay pure filtering.

## Measurements

`p p > t t~`, fixed mu_R = mu_F = 173, lhaid 247000 (NNPDF23_lo_as_0130_qed), mt = 173, 100k events. The integrals agree and the sampling stops being discarded:

| `sqrt_s.min` | before | after | unweighting eff. | samples |
|---|---|---|---|---|
| 900 | 28.073(18) | 28.043(27) | 0.288 → 0.433 | 760k → 628k |
| 6000 | 2.39845(99)e-05 | 2.4000(14)e-05 | 0.036 → 0.493 | 3.49M → 880k |
| 12000 | 5.5789(27)e-14 | 5.5834(27)e-14 | 0.106 → 0.552 | 1.47M → 538k |

The no-op control — a cut at 347, just above the `2 m_t = 346` kinematic threshold, where it can remove almost nothing — reproduces the inclusive 559.22(42) as 558.88(42).

A ladder check on an s-channel-resonance topology (`p p > e+ e-`, lepton cuts relaxed) is self-consistent too: the run at `sqrt_s.min = 100` implies 2.319(53) above 200 GeV and 0.109(11) above 500, against direct runs of 2.3925(33) and 0.10405(15).

## A correction to the record

This was reported as a **~10% cross-section loss** — a cut removing 0.05% of events costing 10.6% of the integral. That does not reproduce, and could not have: the line was dead code, so it could not bias anything. Controlled runs of that exact reproducer give 559.22(42) inclusive and 558.86(42) at `sqrt_s.min = 347` on `main`, and 558.24(39) / 558.34(39) on `feat-mlm`; `sqrt_s.min = 900` gives 28.05, not 20.85. At the mapping level, handing the cut to `PhaseSpaceMapping` before this change left the sampled momenta and weights bit-identical to no cut.

So this PR is an efficiency fix, not a physics fix. No previously published `sqrt_s.min` number is invalidated.

## Test

`madspace/tests/test_sqrt_s_cut.py`, 10 tests. The sharpest assertion uses the `2 m_t` threshold: a cut placed **below** it cannot forbid a single point, so the same random numbers must give back bit-identical momenta and weights — the integral unchanged in the strongest available sense. Alongside it, a phase-space-volume comparison at a cut that removes half the points and 62 sigma of volume, so a boundary that failed to compensate its own Jacobian could not hide in the statistics. Verified to fail on the unpatched build.

Full madspace suite: 1541 passed, 11 skipped, 4 failed — the four pre-existing `test_double_t.py` failures (`ModuleNotFoundError: scipy`), unchanged by this PR.

### DIFF
--- a/madgraph/iolibs/template_files/mg7/run_card.toml
+++ b/madgraph/iolibs/template_files/mg7/run_card.toml
@@ -117,7 +117,10 @@ $multiparticles
 # A pair_mass minimum is also used to bound the integration: where the pair
 # is what a propagator decays into, the cut becomes that propagator's minimum
 # invariant mass, so the forbidden region is never generated rather than
-# generated and thrown away.
+# generated and thrown away. A sqrt_s minimum bounds it the same way, as the
+# minimum partonic energy the luminosity is sampled from, and so does a
+# pair_mass minimum that no single propagator carries. The integral is
+# unaffected either way -- only how much of the sampling is wasted.
 # for all cuts, min or max can be specified
 $cuts
 
--- a/madgraph/iolibs/template_files/mg7/run_card.toml
+++ b/madgraph/iolibs/template_files/mg7/run_card.toml
@@ -114,12 +114,8 @@ $multiparticles
 #      pair_mass is the invariant mass of a pair, taken over every pair the
 #      named groups can form, so "lepton-pair_mass" is the same cut as
 #      "lepton-lepton-sum-mass". sqrt_s is for all outgoing particles)
-# A pair_mass minimum is also used to bound the integration: where the pair
-# is what a propagator decays into, the cut becomes that propagator's minimum
-# invariant mass, so the forbidden region is never generated rather than
-# generated and thrown away. A sqrt_s minimum bounds it the same way, as the
-# minimum partonic energy the luminosity is sampled from, and so does a
-# pair_mass minimum that no single propagator carries. The integral is
+# A pair_mass minimum is also used to bound the integration.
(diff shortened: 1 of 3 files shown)


ramonpeter:
Looks good to me. Thanks for spotting this. But @theoheimel should do the merge if he agrees too.

theoheimel (approved):
looks good

PR #150, opened 2026-09-14
madmatrix FPTYPE=v: custom propagator denominators in double precision

oliviermattelaer:
Follow-up to #108 (now merged into `main`, which this branch includes).

## Problem

With `FPTYPE=v` (denom64: single precision amplitudes, double precision momenta and denominators), `check precision f v` in axial gauge showed `v` on top of `f`.

In axial gauge ALOHA gives massless vector propagators a custom propagator whose denominator is `(P.PBar) P^2`. The madmatrix writer's custom-denominator branch pasted that expression verbatim, and its factors are TMPs computed from the momenta already narrowed to amplitude precision:

```cpp
const cxtype_amp_sv TMP9  = ( P1[0]*P1[0] - P1[1]*P1[1] - P1[2]*P1[2] - P1[3]*P1[3] );  // p^2 in float
const cxtype_denom_sv denom = ... COUP / ( (TMP9 * TMP10) );                          // only the division in double
```

So the cancellation in `E^2 - |p|^2` happened in single precision; the double `dP1` array was declared but unused. Standard propagators (quarks) were not affected.

On the same line, the scalar backend did not compile at all in axial gauge with `FPTYPE=v` (`invalid operands to binary expression ('cxsmpl<double>' and 'cxsmpl<float>')`); SIMD compiled.

## Change

When a routine's custom denominator is made only of momenta-only TMPs, those TMPs, and only those, are computed from `dP` in denominator precision and narrowed once for the numerator; the denominator uses the double versions:

```cpp
const fptype_denom_sv dTMP9 = ( dP1[0]*dP1[0] - dP1[1]*dP1[1] - dP1[2]*dP1[2] - dP1[3]*dP1[3] );
const cxtype_amp_sv TMP9 = fpamp_of_mom( dTMP9 );
const cxtype_denom_sv denom = ... COUP / ( (dTMP9 * dTMP10) );
```

Every other TMP is unchanged, and `dP` is only declared for the legs these TMPs need (also for assembled combined routines, whose dedup now accepts `dTMP`).

## Validation

`check precision v g g > g g g --nb_event=100000` in axial gauge, relative error against `FPTYPE=d`, runs alternated with nothing else on the machine:

| | v mean error | v max error | rate > 1% | d/v speed-up |
|---|---|---|---|---|
| before | 4.674e-06 | 1.052e-01 | 5.0e-05 | 1.57, 1.64 |
| after | 3.652e-07 | 2.636e-05 | 0 | 1.58, 1.61 |

- `f` is unchanged (5.15e-06 / 0.105); axial `v` now matches `v` in unitary gauge (3.70e-07), the single precision amplitude floor.
- Scalar `v` (built by hand, same points): 3.68e-07 mean / 2.14e-05 max.
- All of `d`, `f`, SIMD `v` and scalar `v` compile in axial gauge, with no new warnings.
- The generated `HelAmps_sm.h` for unitary gauge (`p p > e+ e-`) and FD gauge (`u u~ > w+ w-`) is byte-identical to before.
- In axial gauge `d` and `f` move by ~1 ulp and ~1e-7 respectively: the denominator is now a complex divided by a real instead of by a complex.
- A first version also put the numerator-only momenta TMPs in double: same mean error, but ~4% slower `v`, so it was dropped.
- New unit tests in `tests/unit_tests/iolibs/test_madmatrix_writer.py`; together with the precision and language checks, 11/11 pass.

## Notes

- The FD-gauge IOTest golden `FD_madmatrix/src/HelAmps_sm.h` (regenerated on `main` by #108) passes with this change: it does not alter FD output.
- FD gauge's own propagator factor is a separate code path; there `u u~ > w+ w-` gives bit-identical `f` and `v`. Not investigated.


stloufra:
Did not test the `v` with axial gauge. Thanks for the bug fix - can be merged right after the mixed precision.

madmatrix/model_handling.py
```diff
@@ -675,15 +738,39 @@ def define_expression(self):
...
+                        dexpr = self.momenta_in_denom_precision(expr)
+                    if dexpr is not None:
+                        # OM a custom propagator denominator (P.PBar P^2 in axial
+                        # gauge) is computed from the momenta in denominator
+                        # precision (double for FPTYPE=v): the cancellation in
+                        # E^2 - |p|^2 is what loses the digits. Its TMPs are
+                        # narrowed for the numerator, which stays in amplitude
+                        # precision; the other momenta-only TMPs stay as they
+                        # were (computing them in double costs ~4% for no
+                        # measurable gain)
```
stloufra: The comments might be too verbose and keeping the performance benchmarks is redundant in code - lets keep them as artefacts/comments in PRs/issues.

stloufra (approved):

madmatrix/model_handling.py
oliviermattelaer:
```suggestion
                        # tmp variable build from pure momenta
                        #   - single precision is used only in the numerator
                        #   - double precision is used in the denominator
```

PR #152, opened 2026-09-14
Revise LICENSE with new copyright year and authors

ramonpeter:
Updated copyright year and contributors in LICENSE file.

Please check for missing Authors.


oliviermattelaer:
For the year, we should put no year at all. (It has no legal impact and is just annoying to update) Additionally put three single years is quite uncommon.

For the list of authors, I would certainly not remove any names at best some need to move to "former contributor" (like Johan, Michel but certainly not some names which are still active).

Maybe a fair list is also to include all the authors of the MadGraph7 paper (plus former core member). To avoid any political issue, I would close this PR and I would propose that you remove the branch. I will handle the list myself (and when the paper will be out). 
Do not hesitate to add your name in the meantime if you want (but do not remove anyone)

PR #164, opened 2026-09-17
mg7: scale_factor, a factor on the dynamical scale (+ dynamical_scale_choice HT/n)

oliviermattelaer:
## Why

madspace applies **no factor** to its dynamical scale: `EnergyScale`
(`madspace/src/phasespace/scale.cpp`) returns the chosen function itself, and the
run card's `ren_scale` / `fact_scale*` are only read when the scale is *fixed*.
An mg7 run could therefore use H_T (`transverse_mass`) or H_T/2
(`half_transverse_mass`) and nothing else.

H_T/4 — the usual central choice for a high-multiplicity final state such as
four tops — was out of reach, while NLO reaches it with
`dynamical_scale_choice = 3` and `muR_over_ref = muF_over_ref = 0.5`. Putting an
LO and an NLO sample on the *same* scale was impossible, which is what this
started from (a 4-top LO/NLO MadSpin study for the MG7 paper).

`RunCardMG7._legacy_compat` was already reporting a `scalefact` — hard-wired to
1.0.

## What

`[beam] scale_factor` multiplies the **dynamical** scale event by event, with the
same convention as the LO run card's `scalefact`
(`Template/LO/SubProcesses/setscales.f`, *"scale factor for event-by-event
scales"*): a fixed scale is an absolute value and is left alone. So
`half_transverse_mass` with `scale_factor = 0.5` is H_T/4.

- `EnergyScale` gains `scale_factor` (default `1.0`, so every existing card is
  bit-for-bit unchanged) applied to the dynamical branch only.
- `SystematicsConfig` carries it too. The alternative dynamical scales of
  `[systematics] dynamical_scale` are recomputed from the momenta *inside* the
  calculator, so without this they would silently have been the **unscaled**
  H_T / E_T / sqrt(s-hat) while the nominal one carried the factor.
- `RunCardMG7` refuses a non-positive value, and warns when `scale_factor != 1` is
  combined with both scales fixed (where it does nothing) rather than ignoring
  it quietly. `_legacy_compat` reports the real value as `scalefact`, and the LO -> MG7
  card conversion now ports `scalefact` onto `scale_factor` instead of dropping it.
- `set dynamical_scale_choice HT/n` (any n > 0) works the same on the mg7, LO
  and NLO cards and sets the choice together with its factor: mg7
  `half_transverse_mass` + `scale_factor = 2/n`, LO `3` + `scalefact`, NLO `3` +
  `mur_over_ref`/`muf_over_ref` (plain `HT` gives `transverse_mass` / `2`, factor 1).
  So `HT/2` really is H_T/2 even after an earlier `HT/4`. Fixed scales are left
  alone, with a warning. The existing `et`, `shat`, `ckkw` values are unchanged.
- Integer card values containing `/` or `*` that failed to parse became **0**
  without a word, so `set dynamical_scale_choice HT/4` used to select the user
  hook. They now raise `InvalidCmd`, like malformed floats already did (the same
  code is in mg5amcnlo 3.8.1).

## Validation

`p p > t t~`, 20k events, NNPDF40_lo_as_01180, two runs of the same process dir:

| | cross section |
|---|---|
| A `scale_factor = 1.0`, reweighted to `mur = muf = 0.5` | 482.7536 pb |
| B `scale_factor = 0.5`, nominal | 482.8480 pb |
| | **ratio 1.00020** |
| A `scale_factor = 1.0`, nominal | 381.4618 pb |
| B `scale_factor = 0.5`, reweighted to `mur = muf = 2` | 381.5379 pb |
| | **ratio 1.00020** |

The two ways of reaching H_T/4 (and of coming back to H_T/2) agree to 0.02%,
against a 0.26% MC error on each integral. Per event, `SCALUP` in the LHE equals
`scale_factor * H_T/2` to 5e-11 over 3000 events of each run.

New `madspace/tests/test_scale.py` (8 tests) pins the scale definitions
themselves, the multiplication, the default of 1.0, and the fact that a fixed
scale ignores it; `tests/unit_tests/various/test_banner.py` covers the card side.
`madspace/tests/test_systematics.py` still green.

After the rename, end to end through `launch` (`p p > t t~`, same seed):
`set dynamical_scale_choice HT/4` gives 481.21 pb against 481.30 pb from the HT/2
run reweighted to `mur = muf = 0.5`. `test_edit_card.py::test_dynamical_scale_choice_ht`
covers the three run cards, `test_banner.py::test_int_with_operator_is_not_silently_zero`
the parser.

## Not in scope

Separate `mur`/`muf` factors (the NLO `muR_over_ref` / `muF_over_ref` shape).
One factor matches the LO card and the existing `_legacy_compat` field, and
independent variation is already available through `[systematics] mur/muf`.
Easy to extend if wanted.


oliviermattelaer:
Second validation, this time on the process the feature was needed for —
`p p > t t~ t t~` at 13.6 TeV, NNPDF23_nlo_as_0119, **500k unweighted events**
per run, MadNIS on, same seed:

|  | cross section [pb] |
|---|---|
| `scalefact = 1.0` nominal (H_T/2) | 5.299160e-03 |
| `scalefact = 1.0` reweighted to `mur = muf = 0.5` | 8.491983e-03 |
| `scalefact = 0.5` nominal (H_T/4) | 8.491651e-03 |
| `scalefact = 0.5` reweighted to `mur = muf = 2.0` | 5.299158e-03 |

H_T/4 reached directly vs by reweighting: **0.99996**.
H_T/2 recovered from the H_T/4 run vs generated directly: **1.00000**.

(Tighter than the `p p > t t~` numbers in the description because both runs use
the same seed and MadNIS lands on nearly the same sampling; the MC error is
0.023% per integral.)

Worth noting the direction of the effect for a four-top final state: the LO
cross section is **60% larger** at H_T/4 than at H_T/2 (5.30 fb -> 8.49 fb), so
this is not a cosmetic knob — it is the difference between quoting one central
value and another for the whole process.

theoheimel (changes requested):
- Do we want to call it `scale_factor` in the run card for consistency with the snake case that is used in the other settings?
- Do we still need `half_transverse_mass`, since now we can also set `scale_factor` to 0.5?
- I merged #92 with the madspace API documentation. This will also have to be added here (see merge conflicts)
- looks good otherwise

oliviermattelaer:
Done: renamed to `scale_factor` (the LO `scalefact` still maps onto it). Kept `half_transverse_mass`, since it's the choice people use and `scale_factor` combines with it. Docs added after merging #92. Also: `set dynamical_scale_choice HT/n` now sets the choice plus its factor on the mg7/LO/NLO cards.

PR #165, opened 2026-09-17
Tutorials: validate them one at a time, starting with syntax

oliviermattelaer:
Draft, and meant to stay draft until every box below is ticked.

`tutorial list` groups the tutorials into sections, and **Basic** and
**Advanced** are the ones the menu presents as *validated by the developers*.
That claim is enforced rather than decorative: `Tutorial.__init__` refuses an
`ai_generated` tutorial in either of those sections, so a tutorial moves there
only when someone has actually read it end to end.

The goal of this PR is to get every tutorial through that review. Each one is
ticked when it has been walked command by command, its claims checked against
what MG7 really prints, and moved into a validated section.

### Validated

- [x] `lo` — first events at leading order *(Basic)*
- [x] `nlo` — NLO computations with aMC@NLO *(Basic)*
- [x] `model` — working with models *(Advanced)*
- [x] `madloop` — loop matrix elements with MadLoop *(Advanced)*
- [x] `syntax` — the process generation syntax *(Advanced, moved here by this PR)*
- [x] `decays` — decaying unstable particles *(Advanced, moved here by this PR)*

> Depends on #175: `decays` answers a launch card question and then runs MadSpin on a second output, which is exactly the sequence that crashed before it. #175 is merged into this branch so the tutorial can be tested; merge #175 first.
>
> Depends on #182 and #183: `bsm` runs `customize_model` on SMEFTatNLO-NLO, which #182 makes possible (it was refused) and #183 makes start from the `-NLO` restriction (it started again from the full model). Both are merged into this branch; #183 is stacked on #182.

### Still to review

- [ ] `mg7` — tuning the MG7 integrator
- [ ] `madevent` — the MG5-compatible MadEvent path
- [ ] `bsm` — beyond the Standard Model
- [ ] `standalone` — standalone matrix elements
- [ ] `checks` — checking a process before you trust it
- [ ] `exercises` — practise, with the answers checked

---

### `syntax`, reworked

Every step now opens on the counts MG5 just printed for the reader's own
command, and closes on the command it is waiting for. The numbers come from
the session (`_curr_amps` / `total_diagrams()`), never from the text — checked
against a live run for all eight commands, decay chains and accumulated
`add process` included.

- the opening states the coupling orders instead of letting the search choose,
  and says why that matters in a BSM model;
- the interference lesson explains what a `^2` constraint does *not* do: the
  diagrams stay ordinary, so the diagram count can never tell you which term
  you asked for. A **side quest** (`output standalone`, `launch`) shows the
  printed matrix element coming out negative, which no |M|² can be;
- the exclusion lesson invites `$ a` and `$$ a`. Both are `generate` commands
  over the same particles, so they used to count as the lesson's own command
  and jump the reader a step; a **sticky step** now answers them in place;
- the polarised process gives each Z its own decay
  (`p p > z{0} z{T}, (z > e+ e-), (z > mu+ mu-)`), so the two can be told apart
  and there are three frames worth asking for rather than one;
- the tutorial **ends** there with the usual sign-off, and *offers*
  `output madevent` + `launch` to anyone who would rather see `me_frame` being
  chosen than read that it exists. The frame guidance is a `question_hint`, so
  it reaches the reader at the card question;
- cut: the negative-order shorthand, the three interference caveats, the width
  digression and the MadSpin aside — they made the steps too long to find the
  command in.

`me_frame` is a madevent run-card setting today, so that step takes the
madevent path and says so in a way that will age gracefully once madspace
grows the same parameter.

### `session.py`

A sticky step with a callable key is now tried before the plain-key pass. That
is the rule `Step.sticky` already implied — a sticky step must win over the
step it shields — but a callable key could not honour it, because callables
were tried last and a later `generate` step matched first. Needed here because
`generate … $ a` and `generate … / a` share both their lookup keys.

### `misc.py`

`EasterEgg('error')` runs on every failed command and reaches `post_banner`,
which opened `input/authors.md` unconditionally. That file is written by
`bin/create_release.py`, so a checkout has none, and the failure was reported
as a `DEBUG:` line under the user's own error message — three times over for a
command that failed and then quit. A missing `authors.md` is the normal state
of a checkout, not a failure.

### Tests

145 → 150. The new ones: every step ends on its own solution, no two
consecutive steps wait for the same command, the side quest is reachable and
comes back to the main line, and the exclusion extras reach the sticky step
rather than jumping a lesson.

The 3 failures in `TestRealRunNumbers` are pre-existing and environment-only —
they need a built `madspace` for its `380.57(28)` notation, and are red on
`main` in a worktree without one.


oliviermattelaer:
Second pass: the six unvalidated tutorials (excluding `exercises`) now follow
the two rules `syntax` settled on.

1. **every step ends on the command it is waiting for** — no more hunting for
   it in the middle of a thirty-line lesson;
2. **the step that command leads to opens by saying what it produced**, in the
   numbers the reader just watched MG7 print.

The evidence is read back from the session, never written into the text. Two
helpers moved out of `syntax.py` into `session.py` and two more joined them:
`counts_line` (from `_curr_amps`), `model_line` (what `import model` loaded),
`check_line` (the verdict `do_check` leaves in `_comparisons`).

Confirmed against a live run of each tutorial:

| tutorial | what a lesson now opens on |
|---|---|
| `mg7` | `4 processes with 31 diagrams` for `p p > t t~ j`; the output step names the directory it wrote |
| `madevent` | `2 processes with 4 diagrams`, the madevent directory by name; the output step now ends on `launch` instead of a parenthesis about Ctrl-C |
| `bsm` | `MSSM_SLHA2 is loaded: 51 particles, 795 interactions`; the coupling orders read off the model rather than asserted; `2 processes with 20 diagrams`; `2/2 passed, the largest relative difference 3.9e-16` |
| `standalone` | the diagram count; the `--prefix` paragraph moved above the command so the step ends on it |
| `decays` | the count next to plain `p p > t t~`, and the top width read back out of the card the reader just wrote (1.49 GeV here) |
| `checks` | what `import model sm` loaded, and the permutation verdict |

Two things worth knowing:

- `checks` claimed *"Sixteen rows"* for the flavour check. That is right — I
  counted, the summary line says `16/16 passed` — but it read as a property of
  the checker rather than of this process, so it now says so.
- **`lo` breaks rule 1** in three steps (`welcome`, `install madspace`,
  `produce the output`), and it is already in a validated section. I left it
  alone since this pass was scoped to the unvalidated ones — say the word and
  it is a small fix.

Tests are 151, up from 150: the two convention tests now cover every sequenced
tutorial rather than `syntax` alone (with `lo` and `exercises` named as the
exceptions), and a third pins the recap — the step a `generate` leads to has to
quote the count, so a lesson cannot quietly go back to starting from nothing.

The checklist in the description is unchanged: these six follow the
conventions now, but whether they are *validated* is your call, not the test
suite's.

PR #166, opened 2026-09-17
Backend separation 2

stloufra:
Asked Claude to redo the commits in my previous backend separation PR #76  but from the current `main`, so with the current status of MadMatrix. For details refer to the previous PR. 
## Summary

Physically separates backend-owned driver code (cpu / simd / gpu) out of the madmatrix generator's shared and per-process-generated files into `backend/{cpu,simd,gpu}/`, so a given backend's code lives in its own directory instead of being selected at compile time via `#ifdef MGONGPUCPP_GPUIMPL` / `MGONGPU_CPPSIMD`.

## What changed

- **New layout**: `SigmaKin`, `color_sum`, `CrossSectionKernels`, `MatrixElementKernels`, `EventStatistics`, `umami`, the 14 `MemoryAccess*.h` files, and `mgOnGpuConfig/Cxtypes/Fptypes/Vectors.h` now live under `backend/{cpu,simd,gpu}/`, sourced by the Makefile via `-I`/`vpath` instead of being copied per process.
- **New per-process files**: `ColorData.h` (merges colour channel information + the color matrix), `ProcessData.h` (merges contexpr process data e.g. `nwf`, extended with the helicity/flavor tables that used to be inline in `CPPProcess.cc`), `ProcessTables.h` (flavor-coupling / broken-symmetry tables) — single source of truth read by both `CPPProcess` and backend-owned code.
- **Shrunk `CPPProcess.cc`**: all funcs related to `SigmaKin` computation moved to `/backend/Sigmakin` `fpeEnable()`/`getCompiler()` moved to backend files; per-parameter imperative literal-assignment code replaced by a generated pointer-to-member table plus one generic `gatherFptype`/`gatherCxtype` loop written once in `Parameters.h`.
- **Deleted dead code**: `Bridge.h`, `BridgeKernels.*`, `fbridge.*`, `fsampler.cc` (zero remaining references); 
- **`GpuAbstraction.h`/`GpuRuntime.h` made gpu-only.
- **Namespace unification**: `mg5amcCpu`/`mg5amcGpu` → one `madmatrix` namespace (the split is now physical, not namespace-based)
- **Backend call sites, no more `#ifdef` at the point of use**:
  - `createNormalizedColorMatrix()`: no-op inline on cpu/simd, real `gpuMemcpyToSymbol` on gpu.
  - `ColorData.h`'s device→host copy of `channel2iconfig` removed entirely (its only consumer, `runTest.cc`, isn't wired into any export path — kept a local `#ifndef`/`#else` there since it can't get per-process `%`-substituted literals the way `ColorData.h` can for now).
- **Removed `backend/{cpu,gpu}/mgOnGpuVectors.h` entirely** — neither backend has real C++ vector types; their scalar-only content (`neppV`, `fpternary`/`cxternary`/`maskand`, the `_sv` typedefs, `cxabs2`) merged into `mgOnGpuCxtypes.h`. The handful of shared, non-per-backend `.inc` files that need this now pick the right header via `__has_include`. Stripped the remaining always-true `#ifdef MGONGPU_CPPSIMD` guards (and the dead code blocks they wrapped) from `backend/simd/mgOnGpuVectors.h`.
- Standalone files are guarded for `output standalone_mg7` only
- `USEBUILDDIR` defaults to 1 for cleaner builds
- `rambo` used in standalone mode for both massive and massless version as well as the random number kernels live in `./src/rambo` and is sourced when `P1_*` is compiled 

## 

Net effect
- defined `MGONGPUCPP_GPUIMPL` now survives only for poping diagnostics while carrying `hip/cuda` diversification and in the standalone specific files (`rambo/check_sa`)
- defined `MGONGPU_CPPSIMD` now survives only in the /backend/simd as carry for the sims lane size and in the `HelAmps_sm.h` polarisation functions for masking paths of conditional code 

## Test
 - [x] Full 5-backend (scalar/sse4/avx2/avx512/cuda) × 3-process (gg_ttxgg, uu_huugg,epem_mupmum) bitwise match on our v100 machine (GCC + GPU) and Apple M4 Pro (Clang) checking 16256 MatrixElement and MeanMatrixElemValue against the pre-refactor baseline commit (1a4f00df67, the brach from commit on main) using check_sa -p -v 256 64 1


oliviermattelaer:
## Code review

Reviewed at `b7faf4cc1`, against merge base `aa9d133fa`.

I regenerated and rebuilt everything in a scratch checkout rather than reading only the diff, and the mechanical part of the split holds up well:

* `g g > t t~` and `p p > t t~` (2 `P*` dirs) generate and build on `scalar` and `simd_128`, including `FPTYPE=f`.
* `e+ e- > mu+ mu-` reproduces the merge-base `MeanMatrixElemValue` **exactly** (`0.013382 +- 0.000708448`), scalar and simd agreeing.
* The new `.check_sa_backend` relink marker works: switching `simd_128` → `scalar` correctly relinks `check_sa.exe` against the scalar libs.
* `test_export_cpp` and `test_madmatrix_writer` pass (12 tests).
* 22 of the 30 `backend/cpu` ↔ `backend/simd` file pairs are byte-identical, which is reassuring — it shows the split really was mechanical and nothing was quietly rewritten.

Three things regress behaviour that works on the merge base, then a few smaller points.

---

## Needs fixing

### 1. `nMF` is undeclared in `CPPProcess.cc` — every merged-flavor model fails to compile

`nMF` used to be declared in `process_function_definitions.inc:111` (`constexpr int nMF = FLV_COUPLING::max_flavor;`), i.e. in scope for the generated `initProc`. It now lives only in `madmatrix::ProcessTables` (`ProcessTables.h:27`), which neither `CPPProcess.h` nor `CPPProcess.cc` includes — but `model_handling.py:2024` (and the `MGONGPU_HARDCODE_PARAM` variant at `2036`) still emits bare `nMF`:

```cpp
int tIPF_partner1[nMF * nIPF];
int tIPF_partner2[nMF * nIPF];
cxtype tIPF_value[nMF * nIPF];
```

Reproducer:

```
import model MSSM_SLHA2
generate u u~ > n1 n1
output standalone <dir>
```

then `make BACKEND=simd_128`:

```
CPPProcess.cc:149:23: error: use of undeclared identifier 'nMF'
... (9 errors, lines 149-157)
```

The identical generate + build on `aa9d133fa` compiles and runs (`MeanMatrixElemValue = 0.000570885`). `nIPF` resolves fine (it comes from `ProcessData.h`); only `nMF` is missing.

Cheapest fix is probably to qualify it in the codegen (`ProcessTables::nMF`, which resolves inside `namespace madmatrix`) and add `#include "ProcessTables.h"` to `process_h.inc` after `Parameters.h`. Worth adding an MSSM (or any `nIPF > 0`) process to whatever set you validate against, since the SM processes in the description all have `nIPF = 0` and cannot catch this.

### 2. `^2` split-order processes now abort at generation

`edit_colordata` (`model_handling.py:2442`) raises for any process with more than one amplitude split order:

```
generate u u~ > t t~ QED^2==2
output standalone <dir>
```

```
INFO: standalone: '_quark _anti_quark > t t~ QED^2==2 @1' has 3 squared-order components
      (QED=0, QED=2, QED=4); keeping QED=2, dropping QED=0, QED=4. The jamps are split
      over 2 amplitude orders and the color sum pairs them; CPU backends only ...
Exception: Split amplitude orders ('^2' constraints with more than one amplitude order)
      are not yet supported by the backend-separated color sum ...
```

This generates fine on the merge base. I understand from the comment that it is a deliberate scope call — `color_sum.cc` is now shared across all `P*` dirs and can no longer carry the `color_sum_splitorders.cc` variant. Two requests rather than a demand to implement it here:

* Please mention the capability removal in the PR description — interference-only samples (`u u~ > t t~ QED^2==2` is the motivating example in `output.py:236-246`) stop working, and right now the description does not say so.
* `check_split_orders()` announces "the color sum pairs them; CPU backends only" immediately before the crash. Please suppress or reword that INFO on the path that is about to raise; as it stands the exporter tells the user the process is supported and then dies.

Happy for the re-implementation itself to be tracked as a follow-up issue.

### 3. `madmatrix/trex.py:279` calls a method that does not exist

```python
self._link_backend_dirs_in_P(matrix_element)
```

`git grep _link_backend_dirs_in_P` returns only this call site — there is no definition anywhere in the tree, so `TREX_ProcessExporter.generate_subprocess_directory` raises `AttributeError` on the first subprocess.

Given the `# for TREX for now no /backend` comments, I assume the intent was to leave TREX on the old layout, in which case deleting the line is the fix. Note TREX will not build either way: `MadtRex/makefiles/cudacpp_runner.mk` still links `BridgeKernels_cpp.o` / `fbridge_cpp.o` and `MadtRex/template_files/rwgt_runner_cc.inc` still `#include "fbridge.h"`, while this PR deletes `Bridge.h`, `BridgeKernels.{cc,h}`, `fbridge.{cc,h}` and `fsampler.cc`. That is fine if TREX is knowingly parked, but it would be good to say so explicitly somewhere.

---

## Potentially worth changing (not blocking)

### 4. `__has_include("mgOnGpuVectors.h")` as backend detection

`process_h.inc:19` and `cpp_model_parameters_h.inc:20` decide "am I the SIMD backend?" by header presence rather than by a macro from `mgOnGpuConfig.h`. `INCFLAGS` contains `-I$(SRC)` (`madmatrix.mk:455`), and `trex.py` copies `mgOnGpuVectors.h` into `src/`, so a `scalar` or `cuda` TREX build would find it and take the SIMD branch — pulling SIMD vector types into a build whose backend directory deliberately has none. A stale `mgOnGpuVectors.h` left in a `P*` or `src/` dir from an older output does the same. A real macro (or an empty `mgOnGpuVectors.h` stub in `backend/cpu` and `backend/gpu`, as you already did for `MemoryAccessVectors.h`) would make this unambiguous.

### 5. `backend/simd/color_sum.cc` dropped `MGONGPU_CPPSIMD` from the mixed-mode guard

Lines 53 and 100 now read `#if defined MGONGPU_FPTYPE_DOUBLE and defined MGONGPU_FPTYPE2_FLOAT`, while `nParityCS` at line 112 in the same file and `nParity` in `SigmaKin.cc:47` both keep the `MGONGPU_CPPSIMD` term. `backend/simd/mgOnGpuConfig.h:217-219` does `#undef MGONGPU_CPPSIMD` when no ISA macro is visible; in that build with `FPTYPE=m`, `nParity` would be 1 while `color_sum_cpu` still writes `MEs` at `ievt0 + neppV`.

Not reachable today — every `SUPPORTED_CPU_BACKENDS` value other than `scalar` forces an ISA flag (`madmatrix.mk:550-600`) — so this is only about the three guards in one file no longer agreeing with each other.

### 6. `os.listdir()` at import time copies whatever is on disk

`output.py:128` and `135` build the `backend/{cpu,simd,gpu}` and `src/rambo` file lists with `sorted(os.listdir(...))` in the class body. Verified: `touch madgraph/iolibs/template_files/madmatrix/backend/cpu/.DS_Store` and that file lands in every generated `<output>/backend/cpu/`. macOS creates those routinely, and any editor backup or build leftover in those directories ships to users. It also means a missing template dir turns into an `OSError` during `import madmatrix.output` rather than a clear exporter error. An explicit list, or a glob with an extension filter, avoids both.

---

## Filed separately

* #168 — `helcolDenominators[1]` is fed from a list over all matrix elements. Pre-existing (`process_function_definitions.inc:921` on `main`), but this PR moves it into `ProcessData.h:27` and drops the `// assume nprocesses == 1 (#272 and #343)` comment that recorded the assumption; please keep a note of it in the new file.

---

## Optional cleanup — your call, happy for all of it to be follow-ups

* **`process_sigmaKin_function.inc` is now dead but still maintained.** `get_sigmaKin_lines` (`model_handling.py:2107`) still renders it on every subprocess and the result is discarded — the generated `CPPProcess.cc` has no `sigmaKin` body. This PR nevertheless edits the template in lockstep with the three `SigmaKin.cc` copies (`processConfig::ndiagrams` → `ProcessData::ndiagrams`), making it a fourth copy that no build compiles. The next real fix will land in three of four copies silently.
* **Other dead code:** `get_matrix_single_process` (`model_handling.py:2330`), `single_process_template` / `process_matrix.inc` (`1873`), and `blas_color_sum_template` / `color_sum_blas.inc` (`1874`) now have no live caller. The BLAS body was hand-copied into `backend/{cpu,simd}/color_sum.cc`, so the `.inc` will drift and a reader cannot tell which is authoritative.
* **Empty stubs:** `backend/cpu/MemoryAccessVectors.h` and `backend/gpu/MemoryAccessVectors.h` are just two `#include`s and an empty namespace.
* **A `backend/common/` would pay for itself.** Given 22/30 byte-identical `cpu`/`simd` pairs, ~4500 lines now exist in triplicate with nothing keeping them in sync. `b7faf4cc1` in this very PR is the illustration: it had to add one `#include "mgOnGpuConfig.h"` to `backend/simd/SigmaKin.cc` that `backend/cpu/SigmaKin.cc` already had. A `backend/common/` on `INCFLAGS` after `backend/<variant>/` (variant overriding only what genuinely differs) keeps the per-backend clarity without the copies. This is the follow-up I would prioritise.
* **Stale comments:** `model_handling.py:2061` credits `get_matrix_single_process` for the color-flow keys, but it is never called — they come from `set_color_flow_lines_cpp` in `get_process_class_definitions` (`1902`). `output.py:141` says the backend sources are "compiled once"; the build log shows them compiled once **per `P*` directory**, which is in fact required since they include the per-process `ProcessData.h` / `ColorData.h`.
* **`fpeEnable()` in `constexpr_math.h`** (`backend/*/constexpr_math.h:336`) — a runtime SIGFPE-trap installer with a `<cfenv>` dependency and a static latch, in a header otherwise about compile-time math. Anyone auditing where FPE traps get enabled has no reason to look there.

---

Overall: the numerical validation is convincing and the structure is a clear improvement over a single all-backend file. Items 1-3 are the ones I would want resolved before merge; the rest can go to issues.

stloufra:
There is some more work needed to be done. As few things changed and just by redoing the commits claude was not able to make everything compile.

stloufra:
Olivier was able to fix all the caveats during the weekend in the #179. Shall we merge both? @Qubitol @oliviermattelaer

stloufra:
Full 5-backend (`scalar/simd_128/simd_256/avx512y/cuda`) × 3-process (gg_ttxgg, uu_huugg, epem_mupmum) **bitwise match** on our v100 machine (GCC + GPU) — all 5 backends — and Apple M4 Pro (Clang) — `scalar/simd_128` — checking all **16384** `MatrixElement` lines and `MeanMatrixElemValue` against current main tip (`ec8adabc6a`) using `check_sa -p -v 256 64 1`. All 21 backend $\times$ process $\times$ machine combinations pass, built with `-fno-fast-math`
- under the default `-ffast-math` build, separate compilations of identical source can disagree at ~1e-8 relative
- CUDA is unaffected (nvcc's fast-math); passes under its normal build

Those digits are probably anyway numerical error induced and only discovered by different order of rounding. As CI passes, after reviews can be merged @Qubitol @oliviermattelaer 


FYR:
[check_backend_regression.py](https://github.com/user-attachments/files/32469259/check_backend_regression.py)

oliviermattelaer (approved):
let's merge !!!

Qubitol (approved):

PR #173, opened 2026-09-18
madspace: hand every child's pair cuts to ColorOrderedMapping

oliviermattelaer:
## Problem

In `color_ordered` mode, `PhaseSpaceMapping` reorders the pairwise `m_inv_min` / `dr_min` cut matrices from outgoing positions into t-channel child order before building `ColorOrderedMapping`. The loop that builds `child_to_out` had a stray `++a;` next to the loop's own increment, so it skipped every other child. The odd children (2nd, 4th, …) kept the "composite" sentinel, and any pair cut involving one of them never reached the sampler. The bug came in with 91ee1cd76.

The fix removes the stray `++a;`.

## Effect of the bug

`ColorOrderedMapping` uses these matrices only as floors on the invariants it samples: set masses, rest-system masses, and the s₂₃ range of the 2→3 peels. The `Cuts` filter is still applied to the weight afterwards (`phasespace.cpp:652`). A missing floor therefore samples a superset of the allowed region: **it costs efficiency, not correctness.** Checked numerically on e⁺e⁻ → d u s c at √s = 1 TeV (leptonic, one flavour per parton so a cut can name one pair), with m(pair) > 400 GeV:

| topology | cut | pass fraction before → after | cut volume (after) vs filter only / RAMBO |
|---|---|---|---|
| {d u} \| {s c} | m(d,u) | 0.840 → 0.913 | 5.384e4 ± 3.1e2 vs 5.369e4 / 5.406e4 |
| {d u} \| {s c} | m(s,c) | 0.174 → 0.222 | 5.389e4 ± 4.8e2 vs 5.410e4 / 5.408e4 |
| {d} \| {u s c} | m(u,s) | 0.501 → 0.513 | 5.398e4 ± 5.3e2 vs 5.443e4 / 5.408e4 |
| single chain | m(u,s) | 0.272 → 0.317 | 5.481e4 ± 8.7e2 vs 5.474e4 / 5.408e4 |

The pass-fraction numbers are from single runs; the m(d,u) and m(s,c) rows were stable across 10 seeds. With the fix, no point with nonzero weight falls below the cut, and no weight is negative.

## Test

New `madspace/tests/test_color_ordered_pair_cuts.py`:

- **`test_pair_cut_reaches_color_ordered_mapping`**: `PhaseSpaceMapping`'s momenta must match a `ColorOrderedMapping` built by hand with the full matrices on the same inputs (≥ 99% of points within 1e-5 GeV), and differ from one built without the bound. Cases cover a set's composite mass, a 2→3 peel, a chain rest system, and a ΔR bound, each on a pair involving an odd child.
- **`test_pair_cut_on_odd_child_raises_efficiency_not_volume`**: handing the cut to the mapping raises the pass fraction by more than 0.03 and leaves the cut volume unchanged (5σ). Across 10 seeds the pulls stayed within ±1.5σ.

All 7 cases fail on origin/main and pass with the fix. The full madspace suite passes on the fixed build (1574 passed). The only failures are 4 `test_double_t` volume tests needing `scipy` and 2 torch-only modules, all of which fail the same way on origin/main locally.

## Related

The single-chain topology also showed rare, extreme weights (max 6.8e29 in 16M points), with or without cuts. That is a separate precision bug in the 2→3 block, which affects the split topologies too. It is fixed independently in #174. The volume test here sticks to the {d u}|{s c} split, which has no 2→3 step.


ramonpeter:
Thanks @oliviermattelaer I will have a look at it.

ramonpeter (approved):

PR #176, opened 2026-09-18
madspace install.py: explicit compile flags win over saved settings under --yes

oliviermattelaer:
## Problem

In `madspace/install.py`, `main()` checked `args.yes` **before** the command-line flags, both for the compile options and for the GPU architectures. So under `--yes`, explicit `--cuda` / `--hip` / `--openblas` / `--simd` / `--debug` / `--cuda-arch` / `--hip-arch` were dropped, and the build fell back to the saved settings or platform defaults.

A scripted GPU build such as

```
python3 madspace/install.py --source --yes --cuda --cuda-arch 80
```

(or `install madspace --source --yes --cuda ...` from MG5) therefore silently produced a CPU-only build, or built for the default arch (75 / gfx900).

## Fix

Each option is now resolved separately:
1. a flag given on the command line wins;
2. otherwise `--yes` reuses the previous source build's saved value;
3. otherwise the platform default applies.

Unchanged behaviour:
- `--yes` alone still reuses the saved settings.
- Flags without `--yes` still describe the whole build: options you leave out take the platform default, not the saved value.

The resolution is factored into `resolve_compile_options()` and `resolve_gpu_arch()`, and the parser into `build_parser()`. `main(argv=None)` now takes an argument list so it can be tested.

Small additions:
- `--cuda-arch` / `--hip-arch` given without the matching `--cuda` / `--hip` now print a warning that they have no effect, instead of being ignored silently.
- The module docstring and the `--yes`, `--cuda-arch` and `--hip-arch` help texts now describe this precedence.

`bootstrap.py` needs no change: it only ever passes `--source --yes [-j N]`.

## Tests

New file `tests/unit_tests/various/test_madspace_install.py` with 17 tests:
- `resolve_compile_options` / `resolve_gpu_arch`: flag vs saved vs default, with and without `--yes`, negative flags, and the legacy `debug` setting.
- `main()` run end to end with pip and cmake stubbed out, checking the generated `-Ccmake.define...` arguments and the saved settings. Nothing is built.

The new `unittest_madspace_install` job in `unittest.yml` runs them, since CI only runs tests that a workflow names.

- `./tests/test_manager.py test_madspace_install -t0` → `Ran 17 tests, OK`
- With the old "`--yes` first" ordering put back, 9 of the 17 fail.


theoheimel (approved):
Can be merged. One thing that we should also consider for the install script is a reset/uninstall option that triggers `rm -rf madspace/install madspace/build` because sometimes a clean rebuild is necessary

oliviermattelaer:
See follow up #189

Issue #168, opened 2026-09-17
madmatrix: helcolDenominators[1] is too small when a P* directory holds more than one matrix element


oliviermattelaer:
## Summary

In the madmatrix (mg7 C++) exporter, `helcolDenominators` is declared with a fixed size of 1, but it is initialised from a list built over **all** matrix elements of the subprocess directory. A `P*` directory carrying two matrix elements therefore emits an array initialiser with two elements for a one-element array, which is a hard compile error.

## Where

On `main`, in `madgraph/iolibs/template_files/madmatrix/process_function_definitions.inc`:

```cpp
constexpr int helcolDenominators[1] = { %(den_factors)s }; // assume nprocesses == 1 (#272 and #343)
```

filled by `madmatrix/model_handling.py`:

```python
den_factors = [str(me.get_denominator_factor()) for me in self.matrix_elements]
replace_dict['den_factors'] = ",".join(den_factors)
```

So for `len(self.matrix_elements) == 2` the generated code is

```cpp
constexpr int helcolDenominators[1] = { 512, 256 };
```

→ `error: excess elements in array initializer`.

## Notes

* This is **pre-existing**, not introduced by #166. That PR relocates the declaration to the new `ProcessData.h:27` and, in doing so, drops the `// assume nprocesses == 1 (#272 and #343)` comment that was the only in-tree record of the assumption. Filing it here so the caveat is not lost.
* The assumption is now in visible tension with its neighbour in `ProcessData.h`, which computes

  ```python
  replace_dict['nproc'] = sum(2 if m.get('has_mirror_process') else 1 for m in self.matrix_elements)
  ```

  i.e. the file states `nproc` can exceed 1 two lines above declaring a `[1]` array that assumes it cannot.
* Every consumer only ever reads `helcolDenominators[0]` (`backend/*/SigmaKin.cc`, `process_sigmaKin_function.inc`), so nothing depends on the array being larger today.

## Possible resolutions

1. Size the array from the data: `constexpr int helcolDenominators[%(nden)d] = { ... }` and have the consumers index it per process — the real fix if a `P*` directory is ever meant to hold more than one matrix element.
2. If one matrix element per `P*` is a genuine invariant of this exporter, assert it in the generator (`assert len(self.matrix_elements) == 1`) instead of relying on the array size to fail at compile time, and restore the explanatory comment.

Either way it should not stay as an undocumented size-1 array fed from a variable-length list.

PR #179, opened 2026-09-18
Review fixes for #166 (backend separation)

oliviermattelaer:
Fixes for the review of #166, as a PR **into `backend_seperation_PR`** so they can be looked at (and accepted or dropped) separately from the split itself. One commit per item, in the order of the review comment; each commit message says what was wrong, how it was reproduced and how the fix was checked, so any of them can be reverted on its own.

> **Update after merging the new `backend_seperation_PR` (`853b8d5`):** the base branch has since taken its own, equivalent fixes for four of the items below (`287d340` nMF, `8ceb613` `<cstring>`, `24ef1a4` the FD-gauge inline helpers, `6ad4f4f` `DeviceAccessJamp2` + the `CPPProcess.h` include). The merge keeps the base's version of each, so those rows (`a2c5a52`, `1e93379` except `gpu/MemoryBuffers.h`, `61201d4`, and part of `4db0133`) no longer show in this PR's diff. What `4db0133` still adds on top of the base is the `dcNGoodHel` jamp stride and the `atomicAdd` on `colAllJamp2s`.

## Needs fixing (review items 1-3), plus three regressions found while fixing them

| Commit | What |
|---|---|
| `a2c5a52` | **(1)** `nMF` back in scope for `initProc`: every merged-flavor model (e.g. MSSM `u u~ > n1 n1`) failed to compile. Now builds and reproduces the merge-base ME. |
| `4f37adb` | **(new)** Color choice goes through the `ncolor_flow` flows again. The split hard-coded `jamp2_sv[ncolor*iParity+icol] += cxabs2(jamp_sv[icol])`, dropping the per-process `jampflow_lines`. For DDM (all-gluon) processes `ncolor != ncolor_flow`: on `g g > g g g` (6 DDM jamps, 24 flows) the split selected 12 distinct flows, one of them 568/1024 times, against 24 on the merge base. **MEs were bit-identical**, so an ME-only comparison cannot see it. The flow lines now go to a generated `ColorFlows.inc` that `SigmaKin.cc` includes. |
| `4db0133` | **(new, GPU)** `backend/gpu` did not compile (`DeviceAccessJamp2::kernelAccessIcolConst` and `#include "CPPProcess.h"` in `MatrixElementKernels.cc` were dropped), and two device-semantics regressions: `calculate_jamps` wrote jamps with `nhel = 1` while `color_sum_kernel` reads the `[2][ncolor][nGoodHel][nevt]` buffer with `nGoodHel` (`nhel` is the color stride), and the `atomicAdd` on the helicity-shared `colAllJamp2s` became a racy `+=`. All restored to the merge base's own lines. |
| `e364c25` | **(2)** Squared split orders restored instead of refused. The backend `.cc` files are compiled once *per P\* directory* against its generated headers, so a compile-time switch suffices (same pattern as `shouldUseBlas`): `ProcessData::nampso/njampso/nsqampso`, pairing tables in `ColorData.h`, `color_sum_cpu_splitorders` (the old template's pair loop) under `if constexpr( nampso > 1 )`, jamp buffers sized `njampso`; GPU `static_assert`s `nampso == 1` like the old `#error`. `color_sum_splitorders.cc` is folded in and removed. |
| `ce3908a` | **(3)** Drop the call to the undefined `_link_backend_dirs_in_P` in `trex.py`. |

## Worth changing (review items 4-6)

| Commit | What |
|---|---|
| `9507875` | **(4)** Select `mgOnGpuVectors.h` on a `MGONGPU_HAS_VECTORS_H` config macro instead of `__has_include`, which a stray `src/mgOnGpuVectors.h` (TREX copies one) turned true for scalar/GPU builds. Reproduced: the stray header broke `BACKEND=scalar`; now it builds and matches. |
| `0ce9731` | **(5)** Restore `MGONGPU_CPPSIMD` in the two mixed-mode guards of `backend/simd/color_sum.cc`, so they agree with `nParityCS`/`nParity`. Latent only; output unchanged. |
| `c501a7d` | **(6)** Copy only regular `.h`/`.cc` files from the backend and `src/rambo` template dirs: a `.DS_Store` placed there used to land in every output. |

## Optional cleanup (review items 8-11, 13-14; 15 partly)

| Commit | What |
|---|---|
| `d76b6b2` | Stop rendering `process_sigmaKin_function.inc` (result was discarded; it was being kept in sync as a 4th copy of `SigmaKin.cc`); remove it and `color_sum_blas_loop.inc`. |
| `6603911` | Remove `get_matrix_single_process`, `jamp_ncolor`, `process_matrix.inc`, `color_sum_blas.inc`: no caller left. |
| `0cbc16f` | Remove the empty `MemoryAccessVectors.h` of `backend/cpu` and `backend/gpu`. |
| `8701fe0` | `backend/common/` for the 9 files byte-identical in all three backends, searched after `backend/<variant>/`. Deliberately no name in both: a quoted `#include` resolves in the including file's directory first, so a common "host" header would shadow a GPU override. The 6 files shared only by cpu and simd therefore stay duplicated. |
| `e4cf21b` | Stale comments (e.g. a `CalculateJamps.cc` that never existed; backend sources described as compiled once when they are compiled per P\* and must be). |
| `06b52cd` | `fpeEnable()` moves from `constexpr_math.h` to its own `backend/common/fpe.h`. |
| `c5db8e9` | Restore the `nprocesses == 1` caveat on `helcolDenominators`; the actual sizing is #168. |

Items 7 (`HRDCOD=1`, already broken on `main`) and 12 (`.gitignore PROC*`) are intentionally not addressed here.

## CI fixes (the 34 checks that fail on #166 as well)

| Commit | What |
|---|---|
| `1e93379` | `#include <cstring>` in `backend/{cpu,simd}/SigmaKin.cc` and `backend/gpu/MemoryBuffers.h`, which call `memcpy`. clang/libc++ pulls it in transitively, GCC/libstdc++ does not, so on the Linux runners every madmatrix build stopped at `SigmaKin.cc:71: error: 'memcpy' was not declared in this scope`. That one error is 33 of the 34 red checks on #166 (every `acceptancetest_mg7_*`, all `check_xsec_processes`, `acceptancetest_30/32`, `unittest_31`). Reproduced locally with `g++-16`; with the fix, `test_check_language_function_epem_aa`, `test_standalone_cpp`, `test_standalone_split_orders_interference` and `test_e_e_collision_mg7` pass under `CXX=g++-16`. |
| `61201d4` | FD gauge: `define_gauge_dir` / `calculate_propagator_factor` are non-template functions defined in `HelAmps_<model>.h` but marked `INLINE`, which is empty unless `HELINL=1`. Now that `CPPProcess.h` includes HelAmps and `SigmaKin.cc` / `umami.cc` include `CPPProcess.h`, the process library failed to link with duplicate symbols (`acceptancetest_uncovered_merged`, hidden behind the `memcpy` error on #166). Marked `inline`; the FD tests pass with clang and GCC and the golden is regenerated for these lines. |
| `fb9bc7d` | Regenerate the `testIO_FDgauge_madmatrix` golden (`HelAmps_sm.h`): the only differences are the two intended ones, the `madmatrix` namespace and the backend-conditional `mgOnGpuVectors.h` include. |

## Validation

* **Against the merge base `aa9d133fa`**, through UMAMI with identical momenta and random numbers (1024 events): `g g > t t~`, `g g > g g g`, `e+ e- > mu+ mu-`, MSSM `u u~ > n1 n1`, `u u~ > u u~ QED^2==0` are **bit-identical** in ME, per-event selected color and selected helicity (simd_128 mixed; scalar and `FPTYPE=d` spot-checked).
* **Split orders**, `FPTYPE=d`: `u u~ > u u~ QED^2<=4` and `u u~ > d d~ g QED^2==2` bit-identical, `u u~ > u u~ QED^2==2` to 9e-16; `u u~ > t t~ QED^2==2` is analytically zero and noise (~1e-15) on both. In mixed precision the interference differs at the float level (~1e-6 relative), as expected from the float color sum under cancellation. At the RAMBO 1000 GeV point the interference is `-5.5828746494657286e-02` against the recorded `-5.5828746494657258e-02`. `test_standalone_split_orders_interference` passes (it cannot on the current branch).
* **Unit suite**: 1872 tests, same two reds as on the unmodified `backend_seperation_PR` (`scipy` missing; an order-dependent MadLoop IOTest that passes alone).
* **GPU: not run on hardware.** No CUDA/HIP toolchain was available. Every `backend/gpu` translation unit was syntax-checked on the host against minimal CUDA/cuBLAS/thrust stubs, differentially: the merge base's GPU code is clean under the same stubs, `backend_seperation_PR` shows the two compile errors above, this branch is clean (and a split-order process hits the new `static_assert`). The stride and `atomicAdd` fixes are device semantics that check cannot see, so **please build and run the cuda backend** before merging.

## Also noted, not fixed here

* TREX still cannot produce output, for reasons already present on `main`: `edit_rwgt_header` looks up `PLUGIN_export_cpp` on `model_handling` (trex.py defines it in its own namespace), and `librex.so` does not compile with Apple clang (`Rex.h:1609`). Candidates for a separate issue.
* Correction to the review comment on #166: it said 22 of the 30 cpu/simd file pairs were byte-identical. The actual count was **15 of 30**. The recommendation does not change.


madgraph/iolibs/template_files/madmatrix/backend/simd/color_sum.cc
```diff
@@ -50,7 +166,7 @@ namespace madmatrix
     // we gain (not a factor 2...) in speed here as we only loop over the up diagonal part of the matrix.
     // Strangely, CUDA is slower instead, so keep the old implementation for the moment.
     fptype_sv deltaMEs = { 0 };
-#if defined MGONGPU_FPTYPE_DOUBLE and defined MGONGPU_FPTYPE2_FLOAT
+#if defined MGONGPU_CPPSIMD and defined MGONGPU_FPTYPE_DOUBLE and defined MGONGPU_FPTYPE2_FLOAT
```
stloufra, line 169: MGONGPU_CPPSIMD should not be needed anywhere as all the backend specific code is anyway living only in appropriate `/backend` folder. Also we can think of more descriptive define as the one I introduced in #108 but for the `m` (color32) mode. 
```
// macro for computing denom twice to fill rest of SIMD lane for rest
#if defined MGONGPU_CPPSIMD and defined MGONGPU_FPTYPE_FLOAT and defined MGONGPU_FPTYPE_MOMENTA_DOUBLE
#define MGONGPU_SIMD_DENOM64 1
#endif
```

madgraph/iolibs/template_files/madmatrix/backend/simd/color_sum.cc
```diff
@@ -97,7 +213,7 @@ namespace madmatrix
     // NB: color_sum ADDS |M|^2 for one helicity to the running sum of |M|^2 over helicities for the given event(s)
     fptype_sv& MEs_sv = E_ACCESS::kernelAccess( MEs );
     MEs_sv += deltaMEs; // fix #435
-#if defined MGONGPU_FPTYPE_DOUBLE and defined MGONGPU_FPTYPE2_FLOAT
+#if defined MGONGPU_CPPSIMD and defined MGONGPU_FPTYPE_DOUBLE and defined MGONGPU_FPTYPE2_FLOAT
```
stloufra, line 216: Same here.

stloufra (approved):
I really like the idea of putting the commons in the `backend/common` as that was a caveat discussed but never implemented on my side. Also thank you for the rest of the bug fixes.

PR #180, opened 2026-09-18
madmatrix GPU: give the padding events of a batch valid inputs (fixes HIP illegal memory access)

oliviermattelaer:
## Problem

On AMD GPUs every mg7 run aborted at the start of the survey:

```
terminate called after throwing an instance of 'std::runtime_error'
  what():  GPU error: an illegal memory access was encountered
```

The GPU CI of #177 found it on lemaitre4 (MI300A, gfx942, ROCm 7.2.1), `generate p p > t t~`, `device = ["hip"]` ([run 35358676080](https://github.com/MadGraphTeam/MadGraph7/actions/runs/35358676080), job `pp_ttx`).

## Root cause

The fault is in the generated matrix element, not in madspace. With `AMD_SERIALIZE_KERNEL=3 HIP_LAUNCH_BLOCKING=1`, the error surfaces at `umami.cc:517` in `libmadmatrix_P1_gg_ttx_hip.so`, after `sigmaKin`.

`umami_matrix_element` on GPU rounds the batch up to a multiple of the 256-thread block (`rounded_count`), and `sigmaKin` computes every event of the rounded batch. `copy_inputs`, however, returned early for `i_event >= count`. The padding events kept whatever the fresh `gpuMallocAsync` allocation held, including their flavour index, and `cFlavors[iflavor]` then read far out of bounds.

The CPU path of the same function already sets the padding flavour index to 0. The GPU path did not.

## Fix

In `copy_inputs`, the padding events are now copies of the first event. Every computed event then has valid momenta, random numbers, couplings and flavour, and nothing else changes: only the first `count` events are copied out.

## Tests (pp > t t~, 10k events, NNPDF23_lo_as_0130_qed)

| setup | before | after |
|---|---|---|
| HIP, MI300A, color32 (default) | crash (every run, with and without serialized kernels) | 515.5 / 516.7 pb |
| HIP, MI300A, all64 | — | 511.6 pb |
| CPU on the same node | 511.4 pb | — |
| CUDA, A10 (manneback) | 512.7 ± 1.8 pb (CI) | 513.1 pb |

Each value has a statistical error of about ±2 pb.

To confirm the faulting read, a variant that only zeroes the padding events' flavour index also stops the crash (515.6 pb). CUDA runs the same code; there the fresh memory happened to hold harmless values, so the bug was latent rather than absent.

## Regression test

A CPU test cannot reach this code, and a deterministic test would need to poison GPU memory. The lemaitre4 HIP `pp_ttx` job of the GPU CI (#177) reproduces the crash deterministically without this fix, so that job serves as the regression test.

No IOTest golden includes `umami.cc`, so there is no golden drift.


oliviermattelaer:
Not sure why this PR is not closed since the commit is technically merged already. (looks like merge is a no-op)
So I keep it open such that @theoheimel  can have a look. 
But likely cleaner to just close it

PR #188, opened 2026-09-20
Bugfixes of final and intermediate madspace output

theoheimel:
- Do not activate analysis tool by default when launching mg7 runs.
- Do not automatically revert from numpy to LHE format if postprocessing steps were selected. Instead raise an error to signal to the user that they selected a nonsensical combination of settings.
- Ensure correct clean-up of intermediate per-channel npy files.
- In numpy output mode, write a file `header.lhe` with no events but the LHE header/init information.


oliviermattelaer:
For the first point, I would agree if we offer smart plot by default (So I'm working to update the histogram mode to be more automatic --and check if it can handle scale/pdf uncertainty--).

I also think that we should offer a simple way for the user to decide if those are ON/OFF by default (different use case/user might not like the default in the same way, will be in the same update)

For the second point, I just do not agree on it, I do not see the point for a crash, If the user ask for a shower/... we should make it work (how we make it work we can discuss, but it is pointless to make it crash).

oliviermattelaer (changes requested):
The main point is to put back (in one way or another) an automatic switch to lhe when asking for shower or MadAnalysis5, (MadSpin?), ...

oliviermattelaer:
Heads-up: #190 is stacked on this branch (`fix-output-formats` is its base), so it should go in after this one.

It touches two things from here:

- **the LHE auto-revert is put back** — `check_lhe_output_required` raising becomes `force_lhe_output_if_needed` rewriting `output_format` again, on the grounds that asking for a shower in the launch question should be enough without knowing it constrains the output format;
- **the `analysis` switch stays off by default**, but the ExRoot behaviour this PR removed gets a proper home: `input/default_switch.txt`, shared by the LO, NLO and mg7 launch questions, so a user who wants MadAnalysis5 by default can say so once.

The rest of #190 is the mg7 `[histograms]` defaults, the per-run plots with scale/PDF bands, and an opt-in `MADatLO.HwU`.

theoheimel:
My issue with the old system with just a warning is that the user will probably overlook it and then be surprised the output format is wrong. That's what happened for me and why I changed it. Since the numpy output only turns on when the user explicitly requests it (which means they want numpy and not LHE), I don't think it should be quietly changed back. So at least in non-interactive mode it should be a crash.

oliviermattelaer:
Should not we put npy as the default output (if not shower/...)

theoheimel:
Yes, we could consider that. I find it much nicer to use to make plots, inspect the data, etc.

oliviermattelaer:
Superseded by #195, which merges this branch in (header.lhe, ExRoot off by default, per-channel npy cleanup). One deliberate change: the automatic switch to LHE output is kept instead of raising an error, since npy is now the default. header.lhe also replaced the lhe_meta.json #195 had added, and the gridpack header is now complete.

